PR #111, opened 2014-12-15
Upgrade to Django 1.7.2

mhl:
This commit upgrades the project to Django 1.7; the changes here are:

 - admin.autodiscover() is no longer required

 - This switches from South to the new core migrations (among other
   things, this means that syncdb is no longer required as a separate
   step)

 - Upgrade nose and django-nose to avoid a deprecation warning about
   doctest versions:

      RemovedInDjango18Warning: The django.test._doctest module is deprecated;
      use the doctest module from the Python standard library instead.

One of the motivations for this is that we need a bug fix in allauth 0.19.0 but
this bug https://github.com/pennersr/django-allauth/issues/804 means that
migrations don't work with allauth 0.19.0 and Django 1.6. Upgrading to Django
1.7 side-steps that.


.travis.yml
```diff
@@ -29,7 +29,7 @@ install:
     conf/general.yml-example > conf/general.yml
 
 before_script:
-  - ./manage.py syncdb --noinput
+  - ./manage.py migrate --noinput
```
dracos: As the test running creates its own database from nothing, is this line needed at all? (I think I removed it in cb0ca0799caba2f4ab352f94de278f977255fcee anyway.)

mhl: Does SOUTH_TESTS_MIGRATE still work with the 1.7 migrations? (I was assuming not.)

dracos: No, but it still creates an entire new database from scratch, so whether your actual db has run migrate or not doesn't matter AIUI :)

mhl: OK, thanks - indeed, Travis is happy without that, so I've added a commit to remove it.

dracos:
:rabbit: Surprised `autodiscover` was the only change needed :)
Do you want to include https://github.com/mysociety/yournextmp-popit/pull/110#issuecomment-66882229?

mhl:
@dracos I've added that too, thanks - I'll deploy this now.

PR #127, opened 2015-01-07
Remove brackets around candidate party on constituency page

### LINKED ISSUE #124: Too many brackets on constituency page
Look at all these brackets.

screen shot 2015-01-06 at 10 29 47

Making a note, so I can fix this if/when I get the chance. Possible solution is to remove the brackets and put the party name in a lighter grey (so it's closer to the weight of the person's name). Separating the party from the person with a dash might help break it up too.

zarino:
Fixes #124.

Makes candidate parties a bit easier to parse – especially for UKIP candidates who already have brackets in the party name.

screen shot 2015-01-07 at 22 23 01

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,3 +115,10 @@ the development server again are:
 
     cd yournextmp-popit
     ./manage.py runserver 0.0.0.0:8000
+
+### Running the tests
+
+SSH into the vagrant machine, then run:
+
+    cd yournextmp-popit
+    ./manage.py test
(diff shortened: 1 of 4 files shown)


tfgg:
This seems to have broken the candidates.tests.test_constituency_view.TestConstituencyDetailView test. Looks like they just need the expected HTML to be updated.

zarino:
Oh! I didn't realise there *were* any tests. I've fixed the error now, and added a note to the Readme about how to run the tests.

mhl:
:+1: Thanks, @zarino

PR #134, opened 2015-01-09
Fixes for running behind SSL/TLS

mhl:


dracos:
:+1: I think you also need `ACCOUNT_DEFAULT_HTTP_PROTOCOL` just for the resest password form apparently: http://stackoverflow.com/questions/25824598/django-allauth-not-sending-links-with-https - and other email in fixed version of allauth.

mhl:
Thanks, @dracos, I've added that setting too when NGINX_SSL is truthy.

dracos:
I've put a fix for the test failure (due to change in 0.19) in the `only-install-django-once` branch (which also only installs Django once :) ).

PR #139, opened 2015-01-13
Add a privacy policy for YourNextMP

### LINKED ISSUE #106: add a privacy policy statement
Of a simple kind, like TheyWorkForYou's: http://www.theyworkforyou.com/help/#privacy

mhl:
Fixes #106

### DIFF
--- a/candidates/templates/candidates/about.html
+++ b/candidates/templates/candidates/about.html
@@ -42,6 +42,9 @@ after <a href="https://github.com/mysociety/yournextmp-popit/issues/62">further
 discussion</a>) <a href="http://creativecommons.org/publicdomain/zero/1.0/">CC0</a>
 or <a href="http://opendatacommons.org/licenses/odbl/">ODbL</a>.
 
+<p>The privacy policy of this site can be found
+<a href="{% url 'help-privacy' %}">here</a>.</p>
+
 </div>
 
 {% endblock %}
--- a/candidates/templates/candidates/privacy.html
+++ b/candidates/templates/candidates/privacy.html
@@ -1,0 +1,61 @@
+{% extends 'base.html' %}
+
+{% block body_class %}{% endblock %}
+
+{% block title %}Privacy policy{% endblock %}
+
+{% block hero %}
+  <h1>Privacy policy</h1>
+{% endblock %}
+
+{% block content %}
+
+<div class="help-about">
+
+<h3>Simple Version</h3>
+
+<p>This site is run
+by <a href="https://democracyclub.org.uk/">Democracy
+Club</a>.</p>
+
+<p>We make the following promises about your personal
+information as a user of the site:<p>
+
+<ul>
+
+<li>We will not sell or distribute your email address or other
+personal information.</li>
+
+<li>We will gladly show you the personal data we store about you
+in order to run the website.</li>
+
+</ul>
+
+<p>We may use your email address to contact you about your use
+of the site if it's problematic (or particularly helpful!), or
+with occasional announcements about YourNextMP and other
+Democracy Club projects.</p>
+
+<h3>Details</h3>
+
+<p>As well as your login details, we store your IP address along
+with any changes you make to the data on the site.  All this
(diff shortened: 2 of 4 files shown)


candidates/templates/candidates/about.html
```diff
@@ -42,6 +42,43 @@
...
+<h2 id="privacy-policy">Privacy Policy</h2>
+
+<p>If you create an account on YourNextMP.com, then we we will
+store your email address, and log any of your actions that
+change data on the site along with your IP address. If you
+register with a password, then we will store an encrypted
+version of that password, or if you log in via Twitter, Facebook
+or Google, we will store some credentials that enable the login
+to YourNextMP. This personal information will be accessible to
+administrators of the website, who accessible.  We make the
```
symroe: "Who accessible"

symroe:
I think we need to define who "we" are better.  For example, I would like to be able to ask all contributors to YNMP to take part in other Democracy Club projects (there is a high chance that the same people would have an interest in uploading to ElectionLeaflets.org, for example).  If "we" is just the people who made YNMP, this might limit us.

This is one of the harder parts of the Democracy Club "brand"...

candidates/templates/candidates/about.html
```diff
@@ -42,6 +42,43 @@
...
+following promises about all of this personal information:<p>
+
+<ul>
+
+<li>We will not sell or distribute your email address or other
+personal information.</li>
+
+<li>We will gladly show you the personal data we store about you
```
jacksonj04: Are we really "glad" about this? We have no objection to doing it, but I'm guessing SARs don't make us gleeful.

candidates/templates/candidates/about.html
```diff
@@ -42,6 +42,43 @@
...
+
+<p>Please note that this refers to the personal information
+arising from creating a login on the site, not the personal
+information about candidates for general elections - the reason
+for the site to exist is to make information about those
+candidates available to members of the public and website
+developers so that they can find out more about those
+candidates, their promises and policy positions.</p>
+
+<p>Please note that we also use Google Analytics to aggregate
```
jacksonj04: Perhaps worth linking to https://tools.google.com/dlpage/gaoptout/ somewhere in here?

candidates/templates/candidates/about.html
mhl: I took the wording for that bit from TWFY's privacy policy http://www.theyworkforyou.com/help/#privacy - I quite liked the phrasing, since I don't think it's a bad thing to suggest that it's something we're quite happy to do, as well as it being a legal requirement.

mhl:
@symroe Good point about the "we" - I'm rewriting this in response to feedback from mySociety people too, so I'll try to make it clear that it's a project of Democracy Club.

candidates/templates/candidates/about.html
mhl: Good idea, thanks.

candidates/templates/candidates/privacy.html
```diff
@@ -0,0 +1,61 @@
...
+
+<h3>Details</h3>
+
+<p>As well as your login details, we store your IP address along
+with any changes you make to the data on the site.  All this
+information is accessible to administrators of the site.</p>
+
+<p>Please note that this policy refers to the personal
+information arising from creating a login on the site, not the
+personal information about candidates for general elections -
```
symroe: Maybe miss out the word "general" – it's entirely possible this site morphs in to a site that deals with by-elections as well.

mhl: Good point, I've done that.

mhl:
As with my [favourite popcorn recipe](http://www.simplyrecipes.com/recipes/perfect_popcorn/), where you know it's done when the time between pops goes above two seconds, I'm going to assume that that lack of new comments for 10 minutes is a collective :+1: Thanks to everyone for the comments, both here and in IRC.

symroe:
:+1:

PR #154, opened 2015-01-21
More persuasive homepage

### LINKED ISSUE #104: change the message of the homepage
Suggested by @TomSteinberg:

* Emphasize openness and freeness
* If we can get permission add Full Fact / Democracy Club / Unlock Democracy logos / message of support
* Make the message more like: ‘let’s not all do this again and again and again’ - ‘Together we can build a shared database of candidates for 2015 - and save everyone lots of time and money in the process’
* Made by "Democracy Club"

zarino:
Closes #104 – if you agree with the wording @mhl @symroe @ttfg?

This is using a shortened version of @TomSteinberg’s wording from #104. Text can change if you feel it's making unfounded claims.

screen shot 2015-01-21 at 13 08 36

On narrow screens, the layout is largely the same as before, but with the new light cream "description" section inserted below the form and above the leaderboards.


symroe:
Looks lovely!

mhl:
The design looks great, but I think we need to change the text a bit.

Issue #138, opened 2015-01-13
Make the requests for users to help much more prominent and strongly worded


TomSteinberg:
So how about:

"This database of General Election candidates is being built by people like you! Please take a few moments to add extra details about candidates and parties - it only takes a moment."

And also on the candidate pages the call to action is at the bottom, when it should be at the top.

mhl:
We made such a change while the emphasis was on getting people to help with crowd-sourcing, and later the messaging changed to put the emphasis on finding out about candidates instead. This can be closed now.

Issue #142, opened 2015-01-15
Why is candidate gender given such prominence on their profile?


equivalentideas:
Interested in the chosen order of the data on the candidate profile page (for example https://yournextmp.com/person/885/anne-begg).

mhl:
There isn't any particular reason for that: when I added the gender field, the existing fields were (and still are) roughly in order of the complexity of their underlying data storage, so I put it in with the fields that are similarly simple in their Popolo data model representation, like name and email address.

I'm quite happy to reorder these if people think that it appears to put undue prominence on gender? (I'd like to still include that data somewhere, since it's useful for measuring one aspect of diversity.)

equivalentideas:
Hi @mhl thanks for that. Maybe a better order would be by what is easiest to find out, to encourage citizens to dive in and get some quick wins. Gender may be one of the easiest here. I was just interested in the thinking behind this choice.

andylolz:
I think #159 addressed this – gender bumped way down the list.

In reference to @equivalentideas’s last point: The information is not ordered by ease of discovery now, but I think it probably is in order of usefulness. At this stage (now that we have a lot of data in there) that’s probably preferable.

mhl:
@andylolz Yeah, I'm happy that the issue is addressed now, so I'll close this

equivalentideas:
Thanks @andylolz and @mhl , the new profile page layout is very clear, great job.

PR #159, opened 2015-01-22
New profile page layout

### LINKED ISSUE #146: improve the design of the person view page
At the moment, the candidate view page is basically just looks like a table of data:

ynmp-people-view

I thought this was fine, since it suggests "here's a data source you can use" and we didn't necessarily want it to be a destination page to find out about candidates; however, various people have convinced me that this is wrong - we are very high in Google rankings for many candidate-related searches and it's missing an opportunity not providing an engaging page when people follow those links.

Hopefully this should be flexible enough to include links out to social media platforms we haven't included yet (e.g. linkedin, youtube) and potentially allow us to include past election leaflets from electionleaflets.org, etc.

zarino:
Fixes #146.

And partly addresses #138.

profile

### DIFF
--- a/candidates/static/candidates/_content.scss
+++ b/candidates/static/candidates/_content.scss
@@ -1,8 +1,5 @@
 .content {
   background-color: #fff;
-}
-
-.content__inner {
   padding-top: 2em;
   padding-bottom: 2em;
 }
--- a/candidates/static/candidates/_people.scss
+++ b/candidates/static/candidates/_people.scss
@@ -10,6 +10,42 @@
   }
 }
 
+%person-section-heading {
+  line-height: 1.2em;
+  font-size: 1.3em;
+  margin-top: 0;
+  margin-bottom: 0.5em;
+  padding-bottom: 0.5em;
+  border-bottom: 1px solid #ccc;
+}
+
+%responsive-definition-list {
+  // use this to @extend a dl element
+  line-height: 1.4em;
+
+  @media #{$large-up} {
+    dl {
+      @include clearfix();
+    }
+
+    dt {
+      float: left;
+      width: 10em;
+      overflow: hidden;
+      clear: left;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+      margin-bottom: 0.75em;
+    }
+
+    dd {
+      margin-left: 12em;
+      margin-bottom: 0.75em;
+      min-height: 1.4em; // to match dt line-height
+    }
+  }
+}
+
 .candidate-list {
   margin: 1.5em 0;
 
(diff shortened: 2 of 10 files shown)


dracos:
Only thing I'd say is that given #142 it seems to make it even more prominent.

tfgg:
Minor bikeshedding and probably for another time - but is the party too de-emphasized in this new design? It wasn't very prominent beforehand, and now it's quite hard to spot.

mhl:
@zarino Great stuff :) Another thought I had about the gender is that we could have that in a "Demographics" section at the bottom, even if it's the only thing in it. (Although we could put date of birth in that section too.

One other thing, with reference to @tfgg's comment about the display party affiliation - we do have logos for many parties in PopIt, which we can make available to use in the template pretty easily. They're mostly black on a white background, I think, e.g.

* https://yournextmp.popit.mysociety.org/image-proxy/https%3A%2F%2Fyournextmp.popit.mysociety.org%2Forganizations%2Fimages%2F54bea4ae5a9c3744778f9b29%2Fparty%3A90/128/0

Liberal Democrat emblem

struan:
I think the party definitely needs to be more prominent. I'd have assumed it would be in the description under the name - 'Green party candidate for Liverpool, Riverside'

mhl:
I've moved the gender, added a demographics section, changed the wording about the data so it's not emphasizing that we're not offering CSV yet, and also made it more clear what the absence of information for 2010 / 2015 means:

new-profile

I haven't addressed the party prominence issue, but I'll create a new ticket for that, or edit https://github.com/mysociety/yournextmp-popit/issues/162

JoeMitchell:
Looks lovely, well done both. 

Small thing:
- The categorisation of data seems a bit odd. A website / wikipedia page isn't really a Contact Detail. What do you think about recategorising these as: <ul><li>Personal Details,</li> <li>Constituency,</li><li>Contact Details,</li> <li>Web + Social links,</li> <li>Demographics</li>

Prioritisation of data in the page:
- Why do we appear to prioritise their DOB? Isn't having their age in Demographics enough? (Or is it a unique identifier thing?)
- If we work on the assumption that our audience here will tend to be editors adding info, or journos/v interested people (rather than members of the public who are likely to see YNMP data, but dressed up as BBC/Guardian etc)...what's the most important thing to them? I would suggest: Personal Details (Name); Party; Constituency; Email/Phone; Social (starting with Twitter)...

andylolz:
This is looking awesome. V pleased to see you’ve addressed the “absence of information” thing, Mark!

> Why do we appear to prioritise their DOB? Isn't having their age in Demographics enough?

Agreed. I’d remove Personal details completely (since it’s just name & DOB, and name is already written above) and move DOB down to demographics, putting it in grey after their age.

mhl:
Here's a new version taking into account some of those suggestions from @JoeMitchell and @andylolz and incorporating a party logo:

new-profile-view

(The party logo looks a bit silly when it's a thin and tall one, but I'm sure that'd take @zarino no time at all to fix ;))

JoeMitchell:
Boom! I think that looks great. Nice one.

andylolz:
Agree with Joe – This looks great.

btw I didn’t mention the sidebar before – I __really love__ the sidebar.

symroe:
This is incredibly beautiful and exciting :)  Sorry I've not been involved before, and I have no useful input now apart from enthusiasm :)

PR #166, opened 2015-01-27
Strip off any leading @ in Twitter username in validation

mhl:
It's common for people to enter Twitter usernames with a leading '@'
sign, and rejecting this just annoys people - if there's a leading '@',
strip it off and then check that the username only contains valid
characters.


mhl:
(Given a "lgtm" by @andylolz in Slack)

Issue #144, opened 2015-01-15
mark incumbents whose status is still unknown as standing


mhl:
It's looking a bit silly that we're not making the broad assumption that current MPs are standing again - typically we don't have a good citation for these, since people don't tend to write the news story "current MP to stand again!" and only write "current MP to stand down!" stories. (The Conservatives and the Labour Party haven't published canonical lists of current MPs who are restanding either.)

It would be good to have a Django management command which looks for all candidates who meet all of the conditions below:

* They have a parlparse ID in their identifiers
* There's no information about their status in 2015 (i.e. the '2015' key is missing in 'standing_in')
* They're a current MP (i.e. haven't deceased or been replaced in a by-election for any other reasons)

... and mark them as standing in 2015, with a information source of "Incumbent assumed to be standing again".

PR #170, opened 2015-01-28
A script to mark incumbents that we know nothing about as standing

mhl:
This is for issue #144

### DIFF
--- a/candidates/management/commands/candidates_update_incumbents_standing.py
+++ b/candidates/management/commands/candidates_update_incumbents_standing.py
@@ -1,0 +1,97 @@
+from candidates.popit import PopItApiMixin, popit_unwrap_pagination
+
+from django.core.management.base import BaseCommand
+
+from ...update import PersonParseMixin, PersonUpdateMixin
+from ...views import CandidacyMixin
+from ...models import election_date_2015, membership_covers_date
+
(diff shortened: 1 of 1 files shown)


candidates/management/commands/candidates_update_incumbents_standing.py
```diff
@@ -0,0 +1,97 @@
...
+
+            person_data = self.get_person(person_popit_data['id'])
+
+            cons_id = person_data['standing_in']['2010']['post_id']
+            party_id = person_data['party_memberships']['2010']['id']
+            party_name = person_data['party_memberships']['2010']['name']
+            cons_url = 'https://yournextmp.com/constituency/{0}'.format(cons_id)
+
+            # We're now considering marking the candidate as standing
+            # again in the same consituency as in 2015. Check that
```
chrismytton: Should this be "same consituency as in *2010*"?

chrismytton:
Small query about the text of a comment, but otherwise this looks good! :+1:

candidates/management/commands/candidates_update_incumbents_standing.py
mhl: Oh yes, well spotted. I'll fix that.

PR #174, opened 2015-02-01
Include disclaimers on constituency and candidate pages

symroe:
Not confident about the wording, but this is a reasonable start I think.  Happy for others to improve it.

### DIFF
--- a/candidates/static/candidates/_people.scss
+++ b/candidates/static/candidates/_people.scss
@@ -153,3 +153,15 @@
     color: mix(#fff, $color-teal, 70%);
   }
 }
+
+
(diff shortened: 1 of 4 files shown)


mhl:
Perhaps just add something about how it may be inaccurate as well as incomplete? e.g.
> This data is crowd-sourced and may not be complete (or may contain inaccuracies)—please help us to make it better!

mhl:
@symroe I think we should just close this now, if that's OK; I don't *think* we suffered much from not having those disclaimers (and perhaps gained through having less boilerplate text on the page).

Issue #96, opened 2014-12-06
add CSV export


mhl:
Plenty of people would want to consume the data via a CSV export rather than JSON - the output should be modelled on one of the existing big spreadsheets that people use.

@tfgg and @symroe were talking about this, and it's possible that @tfgg has done something on it already?

abimysoc:
We had an email this morning from someone who could certainly have benefited from this feature, hope it launches soon :+1:

mhl:
@abimysoc It's actually live as of this morning, just not advertised on the front-end yet - if someone needs it, the CSV file for all constituencies is at https://yournextmp.com/media/candidates.csv (license CC-BY-SA)

abimysoc:
:heart:

mhl:
This is mentioned on the "API" page, so I'm going to close this now (although that's not to say we shouldn't promote it more, of course): https://yournextmp.com/help/api

PR #177, opened 2015-02-01
CSV export

symroe:
Initial stab at #96.

The view is working fine, but I don't know the best way to save and serve the file (partly because I don't know about the deployment.

I expect it should use django's `Storage` class to write the file to the correct place, but advice from @mhl would be welcome.

### DIFF
--- a/candidates/models.py
+++ b/candidates/models.py
@@ -4,6 +4,8 @@ import jsonpatch
 import jsonpointer
 import re
 import sys
+from collections import defaultdict
+
 from slugify import slugify
 
 from django.contrib.auth.models import User
@@ -15,6 +17,24 @@ from .static_data import MapItData
 
 form_simple_fields = ('name', 'email', 'birth_date', 'gender')
 preserve_fields = ('identifiers', 'other_names', 'phone')
+
+CSV_ROW_FIELDS = [
+    'name',
+    'id',
+    'party',
+    'constituency',
+    'mapit_url',
+    'twitter_username',
+    'facebook_page_url',
+    'party_ppc_page_url',
+    'gender',
+    'facebook_personal_url',
+    'email',
+    'homepage_url',
+    'wikipedia_url',
+    'birth_date',
+]
+
 
 form_complex_fields_locations = {
     'wikipedia_url': {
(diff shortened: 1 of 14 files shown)


symroe:
The script now just outputs CSV to stdout, leaving the saving of the file up to cron.

This is because @tfgg asked for daily snapshots alongside a more up todate file.  It's easiest if cron deals with this.

The URL for the main file is `media/2015_candidates.csv`, and we'll need to find a nice way to list the archives when they appear.

candidates/models.py
```diff
@@ -351,6 +371,35 @@ def delete_memberships(self):
...
+            person_data['facebook_page_url'],
+            person_data['party_ppc_page_url'],
+            person_data['gender'],
+            person_data['facebook_personal_url'],
+            person_data['email'],
+            person_data['homepage_url'],
+            person_data['wikipedia_url'],
+            person_data['birth_date'],
+        ]
+        return row
```
mhl: I think I'd make this a `dict` with keys from `CSV_ROW_FIELDS` and [using `DictWriter`](https://docs.python.org/2/library/csv.html#csv.DictWriter) to output the CSV to avoid having to make sure that the order of these is exactly right when updating the code and so on.

candidates/models.py
```diff
@@ -351,6 +371,35 @@ def delete_memberships(self):
...
+        """
+
+        person_data = defaultdict(str)
+        person_data.update(self.popit_data['versions'][0]['data'])
+        row = [
+            self.name,
+            self.id,
+            person_data['party_memberships']['2015']['name'],
+            self.standing_in['2015']['name'],
+            self.standing_in['2015']['mapit_url'],
```
mhl: It might be good to include the ONS / GSS code for the constituency as well here, which you could get from:
```python
MapItData.constituencies_2010[self.standing_in['2015']['post_id']['codes']['gss']
```
And I'd probably include the bare MapIt ID / post_id as well as the URL.

candidates/models.py
symroe: Right, I thought that too, however `DictWriter` has a required `fieldnames` argument that is a list of field names.  This leaves us in the same position.  I did think about compiling the whole list of dicts and then using `.keys` on the first one to make this list, but I'm not sure that's any better.

mhl: You could just pass `CSV_ROW_FIELDS` as `fieldnames`, though. (I wasn't suggesting getting rid of `CSV_ROW_FIELDS`.)

mhl:
:+1: Great. @symroe: should I go ahead and add the cron jobs, etc. such that it'll work on our deployment system?

symroe:
Yes please, I can't do that at all.  I will add the daily snapshot links once we have some.  Maybe in a different view linked to from the API page.

Issue #18, opened 2014-11-22
show party logos alongside candidates


mhl:
These party logos are in PopIt now (e.g. see https://yournextmp.popit.mysociety.org/organizations/party:52) so we could display them alongside the candidates. (The image proxy should let you request them in a standard thumbnail size, as well, I think.)

mhl:
What I'd quite like this to be like is:

* Constituency page: have a column with party logos as well as the photos of people (perhaps a bit like a ballot paper, but also with any photo as well)
* Person page: have the party logo shown somewhere as well as the photo

The key point, I think, is that there's space for both a party logo and a photograph regardless of whether we have them - mixing photos and logos might create upset. (We've already experienced from support mail that this is a sensitive topic.)

mhl:
I think I'm going to close this, for the reasons given in the linked pull request, but please reopen it if you disagree

PR #188, opened 2015-02-06
Add party logos to the constituency view page (2015 candidates only)

mhl:
This is the start of a fix for #18

The idea is to make this look a bit more like the ballot paper will.

logos-on-constituency-page

... but the alignment's not quite right, and it looks a bit broken on mobile:

party-logos-mobile-broken


mhl:
After thinking about this and talking with @zarino, I think this just makes the constituency page look noisy. The party name is already on the constituency page, and the party logo is very prominent on the individual person page, so I'm doing to ditch this idea and close the bug.

PR #191, opened 2015-02-07
Add middleware to handle errors better if PopIt is down

### LINKED ISSUE #126: Display something nice when popit is down
If popit is down, YNMP just displays 'Server Error (500)'.  We should display something nicer.

mhl:
If PopIt is being restarted, then we currently get a flurry of error
emails and the users hitting the site then see an ugly Internal Server
Error page. This middleware catches ConnectionError (from requests)
and HttpServerError 503 (from slumber) and presents an error page in the
site's style saying the service is temporarily down.

PopIt and YourNextMP are being indepedently monitored by PopIt, so
hopefully this shouldn't affect our awareness of the site being down.

Fixes #126

### DIFF
--- a/candidates/middleware.py
+++ b/candidates/middleware.py
@@ -1,0 +1,16 @@
+from requests.adapters import ConnectionError
+from slumber.exceptions import HttpServerError
+
+from django.shortcuts import render
+
+class PopItDownMiddleware(object):
+
+    def process_exception(self, request, exc):
+        popit_down = False
+        if isinstance(exc, ConnectionError):
+            popit_down = True
+        elif isinstance(exc, HttpServerError) and '503' in unicode(exc):
+            popit_down = True
+        if popit_down:
+            return render(request, 'candidates/popit_down.html', status=503)
+        return None
--- a/mysite/settings.py
+++ b/mysite/settings.py
@@ -96,6 +96,7 @@ INSTALLED_APPS = (
 SITE_ID = 1
 
 MIDDLEWARE_CLASSES = (
+    'candidates.middleware.PopItDownMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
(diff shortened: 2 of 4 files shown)


mhl:
The error page looks like:

popit-down-error

candidates/middleware.py
```diff
@@ -0,0 +1,16 @@
...
+class PopItDownMiddleware(object):
+
+    def process_exception(self, request, exc):
+        popit_down = False
+        if isinstance(exc, ConnectionError):
+            popit_down = True
+        elif isinstance(exc, HttpServerError) and '503' in unicode(exc):
+            popit_down = True
+        if popit_down:
+            return render(request, 'candidates/popit_down.html', status=503)
```
symroe: Should we log the fact it's down anywhere?

mysite/settings.py
```diff
@@ -102,6 +102,7 @@
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'candidates.middleware.PopItDownMiddleware',
```
symroe: Does the ordering matter here?  If so, should we add it earlier on?  Not much point in processing any of the other middlewares if popit is down

candidates/middleware.py
mhl: nagios should notice that YourNextMP is returning 503s (it's monitoring PopIt as well) and this 503 should go in the usual server logs, so I'm not sure if it's worth adding anything additional?

mhl: I'll amend the commit where i said "monitored by PopIt" where I meant "monitored by Nagios" :)

mysite/settings.py
mhl: OK, I'll move it to the top - I don't think that should create any of the problems warned about [here](https://docs.djangoproject.com/en/1.7/ref/middleware/#middleware-ordering)

candidates/middleware.py
symroe: Ah ok, I wasn't aware of how the server was set up.  As long as it's possible to tell how much the site was down at a later date (not actually sure why, but it's good to keep this sort of info somewhere)

mhl: We should be able to recover that (to ~5 minute accuracy) from Nagios (and those alerts also go to an IRC channel)

PR #192, opened 2015-02-07
A basic party detail page and party list page

mhl:
I'd like to deploy these and initially just have the administrators using them, and
see how that goes. The pages are obviously weak design-wise, but hopefully
useful for finding which last seats don't have candidates for particular parties:

example-party-detail-page


chrismytton:
Looks good to me! :+1:

PR #206, opened 2015-02-16
Incomplete field lists and tasks page

symroe:
App for listing incomplete fields.

I wanted to do this in a new app to make things a big cleaner, and also we might add to it in future with more complex views.

Features:

* Pass the field name in to the URL.  This won't work in some cases where the data is only available in a 'deep field' in the returned data.  See `get_field` for some fixes to this
* Try to load a template called `field_[name].html` for the fields, reverting to `field.html`.  This is so we can add advise per field if we want to (like, not adding email addresses that aren't public, etc)
* We don't need to list many records per page, as we show the number at the top of the page.

Problems:

* Searches that search fields in the `versions` array might return incorrect data, as I can't find a way to search in `versions[0]` only.  @mhl do you know how I might do this?
* No presentation on the site.  I'd like to have 'drives' where we put a banner on the site linking to a single field's view.  I'd like feedback on this.

### DIFF
--- a/cached_counts/tests.py
+++ b/cached_counts/tests.py
@@ -6,24 +6,29 @@ from candidates.tests.test_create_person import mock_create_person
 
 from .models import CachedCount
 
+def create_initial_counts(extra=()):
+    initial_counts = (
+        {
+            'count_type': 'constituency',
+            'name': 'Dulwich and West Norwood',
+            'count': 10,
+            'object_id': '65808'
+        },
+        {
+            'count_type': 'party',
+            'name': 'Labour',
+            'count': 0,
+            'object_id': 'party:53'
+        },
+    )
+    initial_counts = initial_counts + extra
+
+    for count in initial_counts:
+        CachedCount(**count).save()
+
 class CachedCountTechCase(TestCase):
     def setUp(self):
-        initial_counts = (
-            {
-                'count_type': 'constituency',
-                'name': 'Dulwich and West Norwood',
-                'count': 10,
-                'object_id': '65808'
-            },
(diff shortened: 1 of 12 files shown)


symroe:
On the searches:

1. `birth_date` [does not work](https://yournextmp.popit.mysociety.org/api/v0.1/search/persons?q=_missing_:birth_date%20AND%20_exists_:standing_in.2015.post_id)
2. `_missing_:contact_details.type:twitter AND _exists_:standing_in.2015.post_id` [raises an exception](https://yournextmp.popit.mysociety.org/api/v0.1/search/persons?q=_missing_:contact_details.type:twitter%20AND%20_exists_:standing_in.2015.post_id) (not sure why), is that what you meant?
3. Even if some version of using `contact_details.type:twitter`, I found it was returning false negatives: i.e., records have a `null` value at the key `contact_details.type:twitter` meaning they were excluded from a search that is designed to find records will `null` or blank values.
4. Good idea to use the fields in `models.py` and reject the others.  I will implement that now.

symroe:
@mhl Actually, based on the above comment about the searches not working, using `form_complex_fields_locations` really won't work.  If we can solve the search queries in a way that actually uses those fields then I can move over to using it.

symroe:
I've done a bit of a clean up and changed the URL to only accept `[a-z_\.\-]`.

mhl:
With regard to the searches, I've created issues for those bugs, sorry about the frustration...

* https://github.com/mysociety/popit/issues/762
* https://github.com/mysociety/popit/issues/764

PR #207, opened 2015-02-16
"Add email" banners

symroe:
Asks very clearly for missing email addresses.

### DIFF
--- a/candidates/templates/candidates/person-view.html
+++ b/candidates/templates/candidates/person-view.html
@@ -32,7 +32,9 @@
 {% block content %}
 
 <div class="person__details">
-
+    <div class="missing_field">
+        <h4>We don't have an email address for {{ person.name }}, help out by adding one!</h4>
+    </div>
   <h2>Personal details:</h2>
 
   <dl>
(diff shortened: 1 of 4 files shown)


candidates/templates/candidates/person-view.html
```diff
@@ -32,7 +32,9 @@
 {% block content %}
 
 <div class="person__details">
-
+    <div class="missing_field">
+        <h4>We don't have an email address for {{ person.name }}, help out by adding one!</h4>
```
andylolz: Could “help out by adding one!” perhaps link to the update page?

mhl:
Definitely seems worth trying, just a small concern about the CSS for the constituency page, otherwise :+1:

tfgg:
Could there be a link to a pre-made Google search next to every candidate? I've previously found it very useful, e.g. http://yournextmep.com/missing/twitter/

symroe:
@tfgg I thought about that, but I don't know how to add the links in a nice way, without cluttering up the interface.  Maybe we can add a link to some guidance on finding data that can pre-populate the URLs?

Somelike like `/person/5151/mandy-rossi/add_data`, with help per field that we're missing?

PR #212, opened 2015-02-19
Script for adding candidates to a twitter list

symroe:


candidates/management/commands/candidates_update_twitter_list.py
```diff
@@ -0,0 +1,82 @@
...
+        if not all((
+            consumer_key,
+            consumer_secret,
+            access_token_key,
+            access_token_secret,
+            self.list_id,
+            self.list_slug)):
+            raise ValueError('Twitter auth details not found in settings')
+
+        self.api = twitter.Api(
```
mhl: I think calling this `self.api` may potentially create confusing problems, since that will replace the `self.api` created by `PopItApiMixin` - maybe `self.twitter_api` would be better.

candidates/management/commands/candidates_update_twitter_list.py
```diff
@@ -0,0 +1,82 @@
...
+            consumer_key=settings.TWITTER_KEY,
+            consumer_secret=settings.TWITTER_SECRET,
+            access_token_key=settings.TWITTER_TOKEN,
+            access_token_secret=settings.TWITTER_TOKEN_SECRET,
+        )
+
+        self.existing_members = set(
+            self.api.GetListMembers(self.list_id, self.list_slug))
+        self.new_members = self.get_all_people() - self.existing_members
```
mhl: I guess it's deliberate that you're not also removing Twitter accounts that have disappeared from YourNextMP, but isn't that needed now? (e.g. us having a parody Ed Balls account as his Twitter account briefly)

candidates/management/commands/candidates_update_twitter_list.py
```diff
@@ -0,0 +1,82 @@
...
+
+        with open(csv_path) as csv_file:
+            csv_reader = csv.DictReader(csv_file)
+            for line in csv_reader:
+                if line['twitter_username']:
+                    self.existing_people.add(line['twitter_username'])
+
+    def get_from_popit(self):
+        for person_dict in popit_unwrap_pagination(self.api.persons,
+                            embed="membership.organization", per_page=100):
```
mhl: A minor thing, but it'll be quicker if you don't include the `embed` parameter, I don't think any of the membership or organization data is needed for this.

PR #214, opened 2015-02-19
Links to search and twitter on missing fields list

symroe:
As per @tfgg suggestion, add links to edit, search and twitter on missing fields list


mhl:
The page looks a bit busy now, but I guess it's fine for the moment :+1:

symroe:
Actually I just noticed that 'email' is hardcoded on the generic field page.  I think that's ok for now as we only link to that one, but I'll make it more generic later on!

PR #218, opened 2015-02-24
Track banner clicks in google analytics

symroe:


dracos:
The hit to GA might not succeed before the page reloads to the new page; you might want something more like https://github.com/mysociety/fixmystreet/blob/master/templates/web/fixmystreet.com/static/posters.html#L16-L31

mhl:
Yeah, come to think of it we had that problem with tracking clicks on a Mzalendo survey at one point, although that FMS example is nicer (e.g. it deals with ctrl-click). Is the timeout just for making sure the new page is visited even if GA is being slow in responding? (The Mzalendo example is here: https://github.com/mysociety/pombola/blob/1bc6e1bba9dbbf8c98e16d5bcaa5de13bb42ad6c/pombola/kenya/static/js/click-tracking.js )

dracos:
Yes, or not responding at all. Could probably be less than 2 seconds too.

mhl:
Since none of us had time to do this in the remaining day and a half (or whatever) that the banner was up, I'm just going to close this now.

PR #221, opened 2015-02-26
Image upload and moderation queue

mhl:
This pull request adds an option for any user to upload a photo of a candidate
to a hidden moderation queue. There is then a staff-only interface to the
review queue where you can crop the submitted image and choose to accept
or reject it; in either case, an email is sent to the uploading user telling them the
result, and, if it's approved, the photo is uploaded to PopIt.


symroe:
:+1: Amazing work Mark :)

PR #230, opened 2015-03-02
Say "Contesting in 2015" before the election.

dracos:
Also remove superfluous "</dd>" making HTML invalid.


mhl:
Just one small comment, otherwise looks great :+1:

PR #233, opened 2015-03-04
Ask authenticated users to agree to copyright assignment

### LINKED ISSUE #69: add a copyright assignment consent screen for contributors
There's now a note on the "About" page saying that by contributing information you're consenting to it being redistributed under CC-BY-SA (or CC0 or ODbL), but [@evdb suggested](https://github.com/mysociety/yournextmp-popit/issues/62#issuecomment-64931049) that it might be safer to have a consent page that people have to click through to before making edits, like this page from YourNextMP in 2010:

* https://github.com/evdb/YourNextMP.com/blob/master/templates/users/grant_copyright.html

mhl:
As part of our plan to make sure that the licensing of the YourNextMP
database is clear, we've been recommended to ask for copyright
assignment from contributors.

This pull request introduces changes such that authenticated users
won't be able to view pages unless they either:
* Agree to the copyright assignnment
* or Log out

The text of the copyright assignment agreement is currently:

---

### YourNextMP user agreement

Before you carry on to edit data in YourNextMP, we need you to agree that your contributions to this site (with the exception of photo uploads) are owned by Democracy Club Limited. In return, we agree to make the complete database available under an open licence such as CC-BY-SA or remove copyright restrictions by releasing into the public domain (CC0).

❏ **Yes, I agree to assign the copyright of my contributions to YourNextMP (apart from photo uploads) to Democracy Club Limited.**

#### Why are we asking you to agree to this?

When we originally set up YourNextMP for the 2015 general election we didn't get legal advice about the best way to deal with copyright and licensing. It turns out that the way that the user agreement that was worded on the 'About' page had a few problems, and in particular restricted what we could do with the data such that we couldn't license the database to some partners who we thought would make good use of the data but who needed slightly different licensing. The simplest way to ensure that we can make the data as widely available as possible while preventing the hard work of contributors to YourNextMP being co-opted by companies building closed candidate databases is to ask you to assign the copyright of your contributions to Democracy Club.

---

A screenshot of this page is here (with slightly out of date text now):

ynmp-copyright-agreement

Fixes #69

### DIFF
--- a/candidates/forms.py
+++ b/candidates/forms.py
@@ -7,6 +7,7 @@ from django.conf import settings
 from django.core.exceptions import ValidationError
 
 from .mapit import get_wmc_from_postcode, BaseMapItException
+from .models import UserTermsAgreement
 
 class PostcodeForm(forms.Form):
     postcode = forms.CharField(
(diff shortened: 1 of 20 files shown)


mhl:
@symroe pointed out that this would be neater as a context processor rather than a mixin - I only did it as a mixin because originally i was applying to views more selectively

tfgg:
"agree that your contributions to this site" should probably exclude photo uploads, since those are handled separately.

(I already said this in slack, just adding to the PR!)

mhl:
I've reimplemented this as middleware, and incorporated the change @tfgg suggested.

PR #245, opened 2015-03-09
Add DC logo to footer

symroe:
dc-logo


zarino:
Notes from me, earlier, in IRC conversation with @mhl:

* I would possibly give it { display: inline } so that "Built by" goes to the left of the logo rather than above it on a separate line
* and then make the logo a bit smaller

I might have a few spare minutes this evening to make the changes.

symroe:
I had it inline initially, but it looked a bit odd, IMHO.  Happy to try it if you like, let me know if you'd rather I do it.

zarino:
@symroe If you want to do it, go for it! Would be good to see how it looks inline. You'd need to make use of `vertical-align` to get it lined up correctly, I guess.

zarino:
Here's an example of how it looks after 1683e6f479c989fbaea55694eeddb514a9741d00:

screen shot 2015-03-09 at 17 48 43

mhl:
Thanks, @symroe and @zarino - that's deployed now :+1:

symroe:
Thanks @zarino!

PR #248, opened 2015-03-10
Optionally disable submit buttons on first click

### LINKED ISSUE #203: Disable candidate save button on first click
Some editors seem to be double clicking the save button when editing candidates, e.g.

https://yournextmp.com/person/6297

A simple bit of javascript that locks the submit button (for a short time?) after the first click should stop this.

zarino:
Fixes #203.

Just add `data-value-after-click="some string"` to any form element (eg: an input or button) you want to disable after the first click.

screen shot 2015-03-10 at 09 18 33

### DIFF
--- a/candidates/static/js/person_form.js
+++ b/candidates/static/js/person_form.js
@@ -98,8 +98,16 @@ function setUpStandingCheckbox() {
   });
 }
 
+function setUpOneClickSubmitButtons() {
+  $('[data-value-after-click]').on('click', function(){
+    var val = $(this).attr('data-value-after-click');
+    $(this).val(val).prop('disabled', true);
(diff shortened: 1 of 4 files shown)


symroe:
:+1: Looks good to me :)

lizconlan:
Hmm, that seems to block the form submit, maybe changing it to hijack the form submit instead would be better. Possibly something like...

    function setUpOneClickSubmitButtons() {
      $('#person-details').on('submit', function() {
        var button = $("input[type='submit']",$(this));
        var val = button.attr('data-value-after-click');
        if (val) {
          button.val(val);
          button.prop('disabled', true);
        }
      });
    }

PR #259, opened 2015-03-12
change wording in person-view for PPC page link

tfgg:
I've seen a significant number of cases of edits by candidates where they seem to think that "Official party page" means their party homepage and put in the URL of their party's main website.

I've updated the wording to bring it more in line with the person edit form, to say "Party PPC page", which I think is more obvious.


mhl:
:+1: Thanks, @tfgg

PR #261, opened 2015-03-13
2010 candidate csv

tfgg:
Add ability to export 2010 version of CSV.


mhl:
Looks good - could you squash these two commits together, though, since the second is just a fixup of the first?

tfgg:
Squash done.

mhl:
:+1: Thanks, @tfgg - I've merged that now.

PR #275, opened 2015-03-21
Finer-grained permissions for merging candidates and reviewing photos

mhl:
Previously, the only way we could give additional permissions to users on the
site was to make them staff or a superuser; these both need you to put a lot
of trust in the users, since they confer other abilities (e.g. to see other users'
email addresses).

This commit introduces two groups called 'Photo Reviewers' and 'Trusted To
Merge' which we can selectively add users to if we trust them to help out with
those tasks.

### DIFF
--- a/auth_helpers/views.py
+++ b/auth_helpers/views.py
@@ -1,0 +1,29 @@
+from django.contrib.auth.decorators import login_required
+from django.contrib.auth.models import Group
+from django.shortcuts import render
+from django.utils.decorators import method_decorator
+
+
+def user_in_group(user, group_name):
+    if not user.is_authenticated():
+        return False
+    group = Group.objects.get(name=group_name)
+    return group in user.groups.all()
+
+
+class GroupRequiredMixin(object):
+
+    """A mixin that requires the user is a member of a particular group
+
+    You should set 'required_group_name' on the class that uses this
+    mixin to the name of the group that the user must be a member of."""
+
+    permission_denied_template = 'auth_helpers/group_permission_denied.html'
+
+    @method_decorator(login_required)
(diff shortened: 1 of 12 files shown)


auth_helpers/views.py
```diff
@@ -0,0 +1,35 @@
...
+class GroupRequiredMixin(object):
+
+    """A mixin that requires the user is a member of a particular group
+
+    You should set 'required_group_name' on the class that uses this
+    mixin to the name of the group that the user must be a member of."""
+
+    permission_denied_template = 'auth_helpers/group_permission_denied.html'
+
+    def dispatch(self, request, *args, **kwargs):
```
dracos: Could you replace the first five lines of this function by decorating this `dispatch` with `@method_decorator(login_required)`?

auth_helpers/views.py
```diff
@@ -0,0 +1,35 @@
...
+        login_url = reverse('account_login')
+        next_path = request.path
+        login_url += '?next=' + urlquote(next_path)
+        if not request.user.is_authenticated():
+            return HttpResponseRedirect(login_url)
+        required_group = Group.objects.get(name=self.required_group_name)
+        if required_group not in request.user.groups.all():
```
dracos: And could these two lines become, now the function exists, `if not user_in_group(request.user, self.required_group_name):`? :)

auth_helpers/views.py
mhl: Yes, good point - thanks.

auth_helpers/views.py
mhl: Oh yes - making it quite obvious what order this code was written in, I suppose... :)

PR #278, opened 2015-03-24
Add electoral commission ID to person dict

symroe:


mhl:
Looks fine other than where the column is added in the CSV output

symroe:
Ordering changed!

PR #280, opened 2015-03-27
Add notification of number photos for review

mhl:
Like this:

photo-review-queue-notification

As well as a reminder for how many photos there are still
for review, this provides a link to the review queue on every
page for users who are able to do something about that.


dracos:
Woo

PR #281, opened 2015-03-27
"Flash" results of photo review actions

mhl:
Previously, when you reviewed a photo you were just redirected
to the photo review list page, with no indication that what you did
just worked. (Also, there was no link to the candidate you'd just
updated, in case you wanted to check that it had worked.)

This pull request uses the Django messages framework to display
"flash" messages of this kind (to use the Rails term). e.g. after
successfully reviewing a photo, you might see:

photo-review-flash

### DIFF
--- a/mysite/templates/base.html
+++ b/mysite/templates/base.html
@@ -78,13 +78,15 @@
       {% if messages %}
         <div class="messages">
           {% for message in messages %}
-            <p class="message {% if message.tags %}{{ message.tags }}{% endif %}">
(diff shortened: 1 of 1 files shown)


dracos:
I think the first commit has a c-n-p mistake in its message. I'd probably merge the 3rd+4th (and possibly) 5th commits together.

mhl:
Yeah, the first commit has just appeared because I noticed the typo in that commit (at the tip of master) and force-pushed a correction :) I'll squash  8436f73 and dd89d00

PR #282, opened 2015-03-27
Add a "make this the primary image" option when reviewing a photo

### LINKED ISSUE #250: Use the most recently added photo
We currently use the earliest uploaded photo, rather than the most recent. The workaround when adding a new photo involves deleting the previous photo (which incidentally causes PopIt to crash – see: mysociety/popit#783).

This is particularly a problem with the new moderation queue, because it’s easy not to realise that the candidate already has a photo. If this is missed, the new photo won’t be used. One workaround would be to flag to the moderator that a candidate already has a photo e.g. by showing it.

mhl:
Sometimes it's clear that the uploaded image is a better one to
use for the candidate, while sometimes it's just an extra one that
it might be nice to have.  Previously, all approved photos were
appended to the images array, which meant they wouldn't be used.
This commit changes that, so that now there's a "make this the
primary image" option, which, when selected, will insert that image
at the front of the images array.

Fixes  #250

### DIFF
--- a/moderation_queue/views.py
+++ b/moderation_queue/views.py
@@ -136,6 +136,7 @@ class PhotoReview(GroupRequiredMixin, PersonParseMixin, PersonUpdateMixin, Templ
                 'y_max': self.queued_image.image.height - 1,
                 'decision': self.queued_image.decision,
                 'moderator_why_allowed': self.queued_image.why_allowed,
+                'make_primary': True,
             }
         )
         context['why_allowed'] = self.queued_image.why_allowed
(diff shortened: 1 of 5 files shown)


moderation_queue/views.py
```diff
@@ -174,13 +175,16 @@ def crop_and_upload_image_to_popit(self, image_filename, crop_bounds, moderator_
...
+            data['index'] = 'first'
         with open(ntf.name) as f:
-            requests.post(
+            result = requests.post(
                 image_upload_url,
                 data=data,
                 files={'image': f.read()},
                 headers={'APIKey': self.api.api_key}
             )
+            print "got result.content", repr(result.content)
```
dracos: This a debug line?

dracos:
Looks good to me

moderation_queue/views.py
mhl: Oops, yes

PR #289, opened 2015-03-30
Add Cache-Control headers with max-age 20 of minutes

mhl:
The Django cache middleware also adds a Vary: Cookie header to
responses, so this should mean that Varnish can serve cached
versions of these pages for 20 minutes to client that aren't
sending cookies.

### DIFF
--- a/mysite/settings.py
+++ b/mysite/settings.py
@@ -107,6 +107,7 @@ INSTALLED_APPS = (
 SITE_ID = 1
 
 MIDDLEWARE_CLASSES = (
+    'django.middleware.cache.UpdateCacheMiddleware',
     'debug_toolbar.middleware.DebugToolbarMiddleware',
     'candidates.middleware.PopItDownMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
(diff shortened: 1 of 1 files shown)


mysite/settings.py
```diff
@@ -300,3 +302,14 @@
 if conf.get('NGINX_SSL'):
     SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
     ACCOUNT_DEFAULT_HTTP_PROTOCOL = 'https'
+
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
```
symroe: The dummy cache is just a no-op cache I think.  Do you know what's installed on the server?

mhl: There's memcache on the server, but for the moment all I want is to set the Cache-Control header to help Varnish. I think switching to doing caching with memcache is too big a change for a Monday evening out-of-hours. Does that make sense?

symroe: Oh I see, ok!

PR #294, opened 2015-04-01
Fix the "fix data" commands

mhl:
These one-off scripts were introduced to fix problems in that data that
meant that indexing of people in Elasticsearch would fail, but had some
bugs in them. Now that https://github.com/mysociety/popit/issues/814 and
https://github.com/mysociety/popit/issues/794 are resolved, it's worth
fixing these so we can try again.

This is related to https://github.com/mysociety/yournextmp-popit/issues/295


symroe:
I don't feel qualified to review this PR as I don't really understand the complex set of broken things it's fixing...Sorry!

chrismytton:
Looks good to me! :+1:

PR #296, opened 2015-04-01
Allow birth year as an alternative to full dob

### LINKED ISSUE #246: Allow editors to specify birth year for a candidate (rather than full date of birth)
@mhl says:

> PopIt can just have a year (or year + month) in the date of birth field, but we're not allowing that to be entered in YourNextMP at the moment. It might be better to allow that and then say to people who don't want to give their full date of birth “just enter the year and month then”.

I tried entering just the year of birth via the YNMP PopIt instance, but it failed to save with no error message shown.

lizconlan:
Allows the editing form to accept a date of birth or year of birth (the label still says 'Date of birth' but the help text will say 'YYYY-MM-DD or YYYY'.

As a result, age will be displayed as 'n-1 or n' if we've just got a year as we can't determine the exact age with any certainty (similar to Wikipedia behaviour but I can't find an example now that I want one).

Fixes #246

### DIFF
--- a/candidates/templates/candidates/person-view.html
+++ b/candidates/templates/candidates/person-view.html
@@ -96,7 +96,11 @@
     <dt>Age</dt>
     {% with dob=person.date_of_birth %}
       {% if dob %}
-        <dd>{{ person.age }}<small class="dob">(Date of birth: {{ dob|date:"jS F Y" }})</small>
+        {% if dob|length_is:"4" %}
+          <dd>{{ person.age }}<small class="dob">(Year of birth: {{ dob }})</small>
+        {% else %}
+          <dd>{{ person.age }}<small class="dob">(Date of birth: {{ dob|date:"jS F Y" }})</small>
+        {% endif %}
       {% else %}
         <dd>Unknown
       {% endif %}</dd>
--- a/candidates/forms.py
+++ b/candidates/forms.py
@@ -7,6 +7,7 @@ from .static_data import MapItData, PartyData
 from django import forms
 from django.conf import settings
 from django.core.exceptions import ValidationError
+from django_date_extensions.fields import ApproximateDateFormField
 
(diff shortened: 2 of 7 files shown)


mhl:
Looks good, but it would be nice to add some test coverage. e.g. you could update test_update_person.py to set just a year, and test_create_person.py to set a full date.

candidates/templates/candidates/person-view.html
```diff
@@ -93,11 +93,17 @@
   <dl>
     <dt>Gender</dt>
     <dd>{% if person.gender %}{{ person.gender|title }}{% else %}Unknown{% endif %}</dd>
-    <dt>Age</dt>
     {% with dob=person.date_of_birth %}
       {% if dob %}
-        <dd>{{ person.age }}<small class="dob">(Date of birth: {{ dob|date:"jS F Y" }})</small>
+        {% if dob|length_is:"4" %}
+          <dt>(Approximate) Age</dt>
```
mhl: Actually, having seen this on the page now, I think it's probably better without the "(Approximate)"! Sorry :(

lizconlan: @mhl :laughing: no problem, changing it back!

mhl:
With this change it works fine if you set the DOB to something of YYYY format, but if you set it to YYYY-MM-DD then putting and update to a person fails. I think this is because the `str` of an `ApproximateDate` behaves differently from that of a `datetime.datetime.date`. You probably want to use the `repr` of the approximate date here: https://github.com/mysociety/yournextmp-popit/blob/master/candidates/views/people.py#L39 because of https://github.com/dracos/django-date-extensions/blob/master/django_date_extensions/fields.py#L45-L65

mhl:
:+1: Looks good, feel free to merge and deploy it

PR #304, opened 2015-04-07
Cache PopIt results for the person and constituency view pages with Memcached

mhl:
This is slightly ugly because the way the people are handled in the source code
is ugly. Hopefully I can finish refactoring the models, and then this could be improved,
but for the moment, this is probably as clean as I can make it.


lizconlan:
:+1: Looks good, sorry for the delay - struggling to run anything locally at the moment

PR #307, opened 2015-04-07
Remove any dissolved parties from /numbers/parties

### LINKED ISSUE #260: "Parties" totals are broken
On https://yournextmp.com/numbers/parties , it says there are 3 candidates for "Christian Party "Proclaiming Christ's Lordship". However, viewing the list reveals significantly more than 3 candidates.

There seem to be more than 3 standing in 2015, but part of the problem is people like e.g. Edward Strike, who stood for this party in 2010 but is standing for UKIP in 2015.

Gerv

mhl:
We've made an effort recently to get rid of all the parties that are now
dissolved, and we now only consider the current parties when counting
numbers of candidates for each party. However, there will still be rows
in the cached_counts_cachedcount table that refer to now-dissolved
parties.

This commit introduces code to purge any such dissolved parties from the
cached_counts_cachedcount table when cached_counts_create is run.

Fixes #260

### DIFF
--- a/cached_counts/management/commands/cached_counts_create.py
+++ b/cached_counts/management/commands/cached_counts_create.py
@@ -74,6 +74,21 @@ class Command(PopItApiMixin, BaseCommand):
 
             self.add_or_update(obj)
 
+        # Remove any parties that have dissolved but still might have
+        # entries in the database:
+        parties_in_database = set(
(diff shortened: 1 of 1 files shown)


symroe:
Looks good to me!

Issue #187, opened 2015-02-05
Add OpenGraph and twitter card tags


symroe:
As requested by Owen, something like the following would make facebook shares much nicer:

`<meta property="og:image" content="http://ia.media-imdb.com/images/rock.jpg" />`

Also, twitter 'card' tags would make twitter shares nicer.

owenblacker:
* [Facebook Open Graph metadata format](http://ogp.me)
* [Twitter Card metadata format](https://dev.twitter.com/cards/types/summary)

mhl:
@frabcus suggested this primer for adding basic OG and Twitter Card tags:

* http://moz.com/blog/meta-data-templates-123

PR #308, opened 2015-04-07
Add OpenGraph and Twitter tags

lizconlan:
Adds Twitter and OpenGraph metadata to the candidate and constituency pages

Addresses #187


symroe:
This looks great!

mhl:
Nice work. I've put this branch on http://candidates.staging.mysociety.org - the source for the Twitter card looks fine (although annoyingly its card validator won't work if robots.txt disallows it) but Facebook's opengraph debugger suggests some improvements:

og-issue

From that, I think the `og:description` field should be something more like "John Doe is standing for The Made-Up Party in Borchester - find out more on YourNextMP", rather than the photo's source.

It would be nice to add similar cards to:

* Constituency pages
* The home page

I don't think it's worth worrying about anywhere else. Both of them can just use the democracy club logo that's already in the footer (or some resized version of it) as a picture, I'd have thought.

lizconlan:
@mhl I think that the `og:description` is being derived because I failed to supply it, will fix (and alter the `twitter:description` to match)

lizconlan:
The description text has got a bit complicated so I've moved it into a filter, the expected output can now include:

"John Doe is standing as an independent candidate in Made Up Constituency in 2015"
"John Doe stood as for the Something Party in Made Up Constituency in 2010" (where "Something Party" is the party name)
"John Doe is standing for The Something Else Party in Borchester in 2015" (where "The Something Else Party" is the party name)

Might be worth noting that with the current logic, it will continue to read "is standing" until 2016.

lizconlan:
Code should be ok now, might not have got the wording right on first attempt though!

candidates/templates/candidates/constituency.html
```diff
@@ -3,6 +3,25 @@
 {% load staticfiles %}
 {% load exists %}
 
+{% block extra_head %}
+    <!-- Open Graph data -->
+    <meta property="og:title" content="Candidates (PPCs) for {{ constituency_name }} in the UK 2015 General Election" />
+    <meta property="og:type" content="article" />
+    <meta property="og:url" content="{{ request.build_absolute_uri }}" />
+    <meta property="og:image" content="{% static 'img/logo.png' %}" />
```
mhl: I'm afraid that if you try this in the the Facebook OpenGraph debug tool https://developers.facebook.com/tools/debug/og/object/ you'll see that it objects to the image URL just being a path rather than the complete URL. I think you'll have to get the static file's path like this: http://stackoverflow.com/a/19369666/223092 and use `request.build_absolution_url` on the result in the view to get the image's absolute URL.

mhl:
I've added a comment about a problem that the Facebook OpenGraph debug tool would have shown up - after dealing with that, could these new pages with that tool and also the Twitter card validator? (I've removed robots.txt from candidates.staging.mysociety.org so you can use the Twitter tool as well: https://cards-dev.twitter.com/validator)

lizconlan:
\*sheepishly signs up for a Facebook account\*

lizconlan:
will probably reword d3c689f (particularly now I've spotted the typo) prior to merge as the summary line is no longer correct

mhl:
:+1: Looks good :)

andylolz:
This is brilliant! Nice one, @lizconlan

PR #310, opened 2015-04-09
Allow trusted users to lock constituencies (fix the set of people)

mhl:
This creates a new "Trusted to Lock" group; these people are allowed to
lock a constituency. This means that anyone outside the group won't be
able to add or remove a person from that constituency, although people
in the group still can.


chrismytton:
I don't have a YNMP VM to test this on right now, but the code looks good to me. :rocket:

PR #332, opened 2015-04-20
Update Google Analytics for this being on a secondary domain

mhl:
We're shortly going to move the crowd-sourcing interface to
edit.yournextmp.com while a static site is served at yournextmp.com.
This changes the Google Analytics tracking code as advised here:

  https://support.google.com/analytics/answer/1034342?hl=en

... for this site being the secondary domain to yournextmp.com

I'm unclear about whether it would be bad to deploy this before
changing this site to be edit.yournextmp.com (i.e. whether it would
lose us analytics events) but I would hope not.


symroe:
Looks fine, but I really don't know enough about GA to know if / how this actually works... :/

PR #339, opened 2015-04-22
Prevent any new user accounts from being created

mhl:
As we get closer to the election and the site gets more attention, it
presumably is more likely to attract malicious editors. This commit
prevents new accounts from being created, but still allows existing
users to log in.

### DIFF
--- a/conf/general.yml-example
+++ b/conf/general.yml-example
@@ -78,3 +78,8 @@ MEDIA_ROOT: null
 # If this is true, you have to be in the 'Trusted to Rename' group in
 # order to change the name of a candidate:
 RESTRICT_RENAMES: false
+
+# If this is set to false, then no new accounts may be created - you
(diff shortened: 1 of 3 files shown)


andylolz:
This is a good PR to have on standby, but I don’t think we should use it unless it becomes necessary (i.e. if we find we have a problem with vandalism).

mhl:
Yeah, I won't merge or deploy it without a group decision.

symroe:
Will there be a nice message explaining this?  Or will it just error?

symroe:
Code LGTM – the nice message is less important right now.

mhl:
If you try to sign up and it's closed [this template is rendered](https://github.com/pennersr/django-allauth/blob/master/allauth/templates/account/signup_closed.html), which I think is probably fine for the moment.

symroe:
Yes, that's perfect.  :+1:

PR #340, opened 2015-04-22
On party pages, only show every constituency for the biggest parties

mhl:
It didn't make sense to just show a long list of constituencies saying
"no party standing here" for almost all of them, so this commit changes
the party pages so it only shows all constituencies in a country for
parties standing candidates in more than 30% of the seats there.

### DIFF
--- a/candidates/tests/test_party_pages.py
+++ b/candidates/tests/test_party_pages.py
@@ -87,9 +87,12 @@ class TestPartyPages(WebTest):
             u"We don't know of any Labour Party candidates in Northern Ireland so far.",
             unicode(response)
         )
-        # Check there's no David Miliband (since he's not standing in 2015):
-        self.assertIn(
-            u'<li>No candidate in <a href="/constituency/65719/south-shields">South Shields</a>',
+        # Check there's no mention of David Miliband's constituency
+        # (since he's not standing in 2015) and we've not added enough
+        # example candidates to reach the threshold where all
+        # constituencies should be shown:
+        self.assertNotIn(
+            u'South Shields',
             unicode(response)
         )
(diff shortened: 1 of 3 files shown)


candidates/tests/test_party_pages.py
```diff
@@ -87,9 +87,12 @@ def test_single_party_page(self, mock_popit, mock_requests):
             u"We don't know of any Labour Party candidates in Northern Ireland so far.",
             unicode(response)
         )
-        # Check there's no David Miliband (since he's not standing in 2015):
-        self.assertIn(
-            u'<li>No candidate in <a href="/constituency/65719/south-shields">South Shields</a>',
+        # Check there's no mention of David Miliband's constituencie
```
andylolz: “constituencie” should be “constituency” :)

mhl: Thanks, fixup pushed

andylolz:
lgtm! :rocket:

candidates/views/parties.py
```diff
@@ -92,11 +109,14 @@ def get_context_data(self, **kwargs):
...
-                        for c in MapItData.constituencies_2010_by_country[country]
-                    ]
+                constituencies = [
+                    (c[0], c[1], by_country[country].get(c[0]))
+                    for c in MapItData.constituencies_2010_by_country[country]
+                ]
+                candidates_by_country[country] = {
+                    'constituencies': constituencies,
+                    'stats': get_country_stats(constituencies),
+                }
```
andylolz: this whole bit seems a bit bonkers… But this PR doesn’t make it any more bonkers.

PR #343, opened 2015-04-24
Disallow name changes for all but trusted users

mhl:
As we're getting closer to the election, the data is changing much
less and is much more prominent - e.g. the names and parties are
being shown on Google, which will make us more of a target for abuse.

This commit adds a "Trusted to Rename" group, for whom the site will
work as before. For any not in that group, if they try to change the
name of a candidate the update won't be applied and they'll be told
that their attempted name change has been emailed to the volunteers
to be checked and applied, if it looks sound.


mhl:
@symroe gave this a :+1: in Slack

PR #344, opened 2015-04-24
Photo review: use face detection from OpenCV to suggest a crop

mhl:
This seems to work pretty well, but it's quite CPU intensive, so
the review page will take a couple of seconds to load; we may decide
that in practice this isn't worth it after all, but it seems to be
worth trying.


mhl:
Actually, really it would better to add `crop_min_x`, `crop_min_y`, `crop_max_x` and `crop_max_y` as fields to QueuedImage. A cron job running every 10 minutes could check if these are null, and if not, set them from the face detection's best guess. Then when an image is reviewed, they could be updated to the actual crop bounds picked by the user. This has at least two advantages:

* The review page loading won't be slowed down
* It's good to record the crop boundaries anyway

PR #363, opened 2015-05-01
Allow privileged users to set a winner, and display known winners

mhl:
If there's time, we can backport the design changes of @wrightmartin 
and @zarino from YourNextMP-Read to here, but we should merge
this first in any case.


wrightmartin:
@mhl – give me a shout this morning, I'm ready to start the design work on this asap

candidates/templates/candidates/constituency.html
```diff
@@ -74,6 +75,13 @@
     {% endif %}
   {% endif %}
 
+  {% if user_can_record_results and show_retract_result %}
+    <form action="{% url 'retract-winner' mapit_area_id=mapit_area_id %}" method="post">
+      {% csrf_token %}
+      <input type="submit" class="button candidates-list__actions__winner-mistake" value="Fix an incorrectly recorded winner">
```
symroe: Can this be re-worded in a way the says it will only remove the current winner?  As in, to 'fix' an incorrectly reported person, you need to click this, then click on the correct winner

mhl: How about "Unset the recorded winner, if incorrect" or "Unset the current winner, if incorrect"?

PR #368, opened 2015-05-06
Fetch any parlparse (TheyWorkForYou) IDs for newly elected MPs

mhl:
We'd like to link through to the TheyWorkForYou pages for all new MPs
from YourNextMP, but for people who have never been in TheyWorkForYou
previously, we won't have a parlparse / TWFY ID for them. This script
will fetch parlparse's people.json and look for any people in there
who have a yournextmp identifier, and check that we have the correct
parlparse ID for that person, updating it if it was missing.


symroe:
LGTM :+1:

PR #369, opened 2015-05-08
Add an option to turn off edits for non-superusers

mhl:
This is a pretty thoughtless implementation - it just deals with edits
from the update person page, and not merges, photo uploads and so on.
However as a (very tired) measure just to mean that we don't have to
worry about checking the recent changes now the election is over, this
is probably good enough.


candidates/views/people.py
```diff
@@ -1,7 +1,8 @@
-import re
+aimport re
```
chrismytton: This doesn't seem quite right!

chrismytton:
One typo, but otherwise looks good! :+1:

PR #371, opened 2015-05-11
Improvements to Readme and install experience.

martinszy:
I added some pre-requisites for the install in the readme and I added a commented-out line in the vagrant file to ease the debugging by having a visible screen while booting the vagrant vm.


mhl:
I'm very sorry that this was left waiting for review for so long.  I've forward-ported the documentation changes, which are now merged into master as:

* ba126203bad60c789e
* 7f5ce3cef19c
* f264ea9ab62ddcb

(From the Vagrant documentation, I think `v.ui` was a typo which should be `v.gui` so I've changed that.)

PR #377, opened 2015-06-05
Initial support for generic elections

mhl:
The work of making this codebase easily repurposable for other elections isn't finished, but this pull request is already so large it'll be tricky to review, so I'd like to get it reviewed and merged to master before doing further work. Although support for Argentina isn't complete, the UK general election site (YourNextMP) should still work fine, so we should be able to deploy this without breaking that site.

Known things still to do:

* Make generic the `<title>` tags of every page
* Make generic the OpenGraph and Twitter Card metadata
* Cache the geocoder and MapIt results from the new generic address lookup
* Incorporate the Argentine party data, which Martín is working on
* Fix the ConstituencyDetailView for other elections (there's a reference to the WMC type which is UK specific)
* Create "party sets" for provinces in Argentina, as with countries of the UK
* Any remaining uses of ARBITRARY_CURRENT_ELECTION mark points where the code doesn't support multiple elections yet
* Fix the Javascript in `candidates/static/js/person_form.js` to correctly create select2 widgets and dynamically update the party drop-down that's used, based on party sets
* Make all strings used in the interface translatable
* Fix the "add new candidate" buttons in candidate lists when you're not logged in
* The title of some `/areas...` pages are silly at the moment:

bad-titles


bin/switch-country.py
```diff
@@ -0,0 +1,46 @@
...
+    if not os.path.islink(symlink_filename):
+        print >> sys.stderr, "%s was not a symlink, and should be" % (symlink_filename,)
+        sys.exit(1)
+    full_target_filename = os.path.join(os.path.dirname(symlink_filename),
+                                        target_filename)
+    if not os.path.exists(full_target_filename):
+        print >> sys.stderr, "The intended target of the symlink (%s) didn't exist" % (target_filename,)
+        sys.exit(1)
+    os.unlink(symlink_filename)
+    os.symlink(target_filename, symlink_filename)
```
dracos: Could this also `touch` manage.py or something to restart a `runserver` if it's running?

dracos:
Sorry didn't get it all done today, will finish tomorrow. All looks fine so far, well planned and so on, just some very small queries and comments.

dracos:
Done! All looks good, quite a few minor things (proportional to the size of the changes I think, rather than anything else!).

elections/ar_elections_2015/settings.py
```diff
@@ -0,0 +1,74 @@
+# -*- coding: utf-8 -*-
+
+from datetime import date
+
+ELECTIONS = {
+    # Party primaries for deputies.
+
+    # For this and the next two elections in this module, we don't want to
+    # create an organization membership on winning the eleciton, since these
```
dracos: Typo :)

candidates/templates/candidates/person-view.html
```diff
@@ -182,7 +184,9 @@
       <p>Our database is built by people like you.</p>
       <p>Please do add extra details about this candidate – it only takes a moment.</p>
       {% if user.is_authenticated %}
-        <a href="{% url 'person-update' person_id=person.id %}" class="button">Edit candidate</a>
+        <div>
+          <a href="{% url 'person-update' person_id=person.id %}" class="button">Edit candidate for {{ election_data.name }}</a>
```
dracos: Do you have `election_data` here?

bin/switch-country.py
mhl: Nice idea, I've done that in: d6d66883964d42f

PR #390, opened 2015-06-12
Instead of checking cropped image's MD5 hash, just check format and size

mhl:
It seems that depending on the platform (or Pillow's supporting
libraries) the resulting image files won't be identical, so checking the
MD5sum isn't helpful.  Instead, it's probably fine to check that there's
an image file that's a PNG and of the right dimensions.


dracos:
:+1:

PR #391, opened 2015-06-12
Make OpenGraph and Twitter metadata generic

### LINKED ISSUE #383: make the OpenGraph and Twitter Card metadata generic
... i.e. using settings.ELECTIONS and friends rather than any UK-specific logic

lizconlan:
Adds per app `SITE_NAME`, `SITE_DESC `, `SITE_TWITTER` and `SITE_LOCALE` settings which are used in the Twitter and OpenGraph metadata rather than hardcoding "General Election", "en_GB" etc

Also removes "(PPCs) " from the metadata text.

Closes #383

### DIFF
--- a/candidates/templates/candidates/person-view.html
+++ b/candidates/templates/candidates/person-view.html
@@ -7,11 +7,9 @@
 {% load metadescription %}
 
 {% block extra_head %}
-  <link rel="canonical" href="{{ canonical_url }}" />
-
-  {% with last_cons=person.last_cons %}
+    <link rel="canonical" href="{{ canonical_url }}" />
     <!-- Open Graph data -->
-    <meta property="og:title" content="{{ person.name }}{% if last_cons %} - Candidate (PPC) for {{ last_cons.1.name }} in {{ last_cons.0 }}{% endif %}" />
+    <meta property="og:title" content="{{ person.name }}{% if last_election %} - {% blocktrans with post=constituency election=contested_election %}Candidate for {{ post }} in {{ election }}{% endblocktrans %}{% endif %}" />
     <meta property="og:type" content="article" />
     <meta property="og:url" content="{{ canonical_url }}" />
     {% if person.image %}
@@ -23,14 +21,14 @@
     {% endif %}
     <meta property="og:image:height" content="80" />
     <meta property="og:image:width" content="80" />
-    <meta property="og:site_name" content="YourNextMP" />
-    <meta property="og:description" content="{% metadescription person last_cons DATE_TODAY %}">
-    <meta property="og:locale" content="en_GB" />
+    <meta property="og:site_name" content="{% trans "YourNextMP" %}" />
+    <meta property="og:description" content="{% metadescription person person.last_cons DATE_TODAY %}">
+    <meta property="og:locale" content="{{ LOCALE }}" />
 
     <!-- Twitter card data -->
     <meta name="twitter:card" content="summary" />
-    <meta name="twitter:site" content="@yournextmp" />
-    <meta name="twitter:title" content="{{ person.name }}{% if last_cons %} - Candidate (PPC) for {{ last_cons.1.name }} in {{ last_cons.0 }}{% endif %}" />
+    <meta name="twitter:site" content="{% trans "@yournextmp" %}" />
+    <meta name="twitter:title" content="{{ person.name }}{% if last_election %} - {% blocktrans with post=constituency election=contested_election %}Candidate for {{ post }} in {{ election }}{% endblocktrans %}{% endif %}" />
     {% if person.image %}
       <meta property="twitter:image" content="{{ person.image }}" />
     {% elif person.gender|lower == 'female' %}
@@ -41,8 +39,7 @@
     <meta property="twitter:image:width" content="160" />
     <meta property="twitter:image:height" content="160" />
     <meta name="twitter:url" content="{{ canonical_url }}" />
-    <meta name="twitter:description" content="{% metadescription person last_cons DATE_TODAY %}" />
-  {% endwith %}
+    <meta name="twitter:description" content="{% metadescription person person.last_cons DATE_TODAY %}" />
 {% endblock %}
 
 {% block body_class %}person{% endblock %}
--- a/candidates/templatetags/metadescription.py
+++ b/candidates/templatetags/metadescription.py
@@ -4,14 +4,14 @@ from django import template
 from django.conf import settings
 from django.contrib.staticfiles.templatetags.staticfiles import static
 from django.utils.translation import ugettext as _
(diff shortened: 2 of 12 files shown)


candidates/templatetags/metadescription.py
```diff
@@ -19,24 +20,28 @@ def metadescription(person, last_cons, today):
...
-            if last_party_name == "Independent":
+            if last_party_name == _("Independent"):
                 output = _("%(name)s stood as an independent candidate in %(post)s in %(election)s") % args
             else:
                 output = _("%(name)s stood for %(party)s in %(post)s in %(election)s") % args
         else:
-            if last_party_name == "Independent":
-                output = _("%(name)s is standing as an independent candidate in %(post)s in %(election)s") % args
+            if last_party_name.strip() == _("Independent"):
+                output = _("%(name)s is standing as an independent candidate in %(post)s in %(election)s")
```
dracos: This is missing the `% args` at the end.

candidates/templatetags/metadescription.py
```diff
@@ -19,24 +20,28 @@ def metadescription(person, last_cons, today):
...
         if is_post_election(last_cons[0], today):
-            if last_party_name == "Independent":
+            if last_party_name == _("Independent"):
                 output = _("%(name)s stood as an independent candidate in %(post)s in %(election)s") % args
             else:
                 output = _("%(name)s stood for %(party)s in %(post)s in %(election)s") % args
         else:
-            if last_party_name == "Independent":
-                output = _("%(name)s is standing as an independent candidate in %(post)s in %(election)s") % args
+            if last_party_name.strip() == _("Independent"):
```
dracos: Hasn't last_party_name already been stripped by this point?

candidates/templatetags/metadescription.py
```diff
@@ -19,24 +20,28 @@ def metadescription(person, last_cons, today):
...
+                output = _("%(name)s is standing as an independent candidate in %(post)s in %(election)s")
             else:
                 output = _("%(name)s is standing for %(party)s in %(post)s in %(election)s") % args
-    else:
-        output = person.name
```
dracos: Why was the else case taken out?

candidates/templatetags/metadescription.py
```diff
@@ -19,24 +20,28 @@ def metadescription(person, last_cons, today):
...
-
-    output += " - " + _("find out more on YourNextMP")
+    output += _(" - find out more on YourNextMP")
```
dracos: Not sure if it's better to have the hyphen in the translation or not; I think I'd lean towards outside, in case the string is used elsewhere (e.g. in some link text).

candidates/templatetags/metadescription.py
```diff
@@ -12,6 +12,7 @@ def metadescription(person, last_cons, today):
     if person.last_party:
         last_party_name = person.last_party['name'].strip()
         last_party_name = re.sub('^The ', 'the ', last_party_name)
+        last_party_name = format_party_name(last_party_name)
```
dracos: The line or two above this might as well be in the format_party_name function.

candidates/templatetags/metadescription.py
```diff
@@ -19,24 +20,28 @@ def metadescription(person, last_cons, today):
...
+    output += _(" - find out more on YourNextMP")
     return output
 
 def is_post_election(election, today):
     return today > settings.ELECTIONS[election]['election_date']
 
+def format_party_name(party_name):
+    party_words = party_name.split()
+    if party_words[0] != "the":
+        return "the %s" % party_name
```
dracos: This won't work for any non-English language :) Probably easiest if it just checks and only does this for English?

candidates/views/frontpage.py
```diff
@@ -34,4 +34,6 @@ def get_context_data(self, **kwargs):
         context = super(AddressFinderView, self).get_context_data(**kwargs)
         context['top_users'] = self.get_leaderboards()[1]['rows'][:8]
         context['recent_actions'] = self.get_recent_changes_queryset()[:5]
+        for election in settings.ELECTIONS_CURRENT:
+            context['election_data'] = election[1]
```
dracos: If you only care about the last entry, easier to write `context['election_data'] = settings.ELECTIONS_CURRENT[-1][1]`.

candidates/templates/candidates/person-view.html
```diff
@@ -7,11 +7,9 @@
 {% load metadescription %}
 
 {% block extra_head %}
-  <link rel="canonical" href="{{ canonical_url }}" />
-
-  {% with last_cons=person.last_cons %}
+    <link rel="canonical" href="{{ canonical_url }}" />
     <!-- Open Graph data -->
-    <meta property="og:title" content="{{ person.name }}{% if last_cons %} - Candidate (PPC) for {{ last_cons.1.name }} in {{ last_cons.0 }}{% endif %}" />
+    <meta property="og:title" content="{{ person.name }}{% if last_election %}{% blocktrans %} - Candidate for {{ constituency }} in {{ contested_election }}{% endblocktrans %}{% endif %}" />
```
dracos: Ditto with the hyphen outside the translation (and below).

dracos: It'd be great if this translation could be identical to the one elsewhere in the file for the heading: "Candidate for {{ post }} in {{ election }}" so there's only one string to be translated.

dracos:
Some comments, nothing major :)

candidates/templatetags/metadescription.py
lizconlan: in error, good catch thanks!

candidates/views/frontpage.py
```diff
@@ -34,4 +34,5 @@ def get_context_data(self, **kwargs):
         context = super(AddressFinderView, self).get_context_data(**kwargs)
         context['top_users'] = self.get_leaderboards()[1]['rows'][:8]
         context['recent_actions'] = self.get_recent_changes_queryset()[:5]
+        context['election_data'] = settings.ELECTIONS_CURRENT[-1][1]
```
mhl: This file's missing a `from django.conf import settings`

mhl:
I added a note about one small bug.

I've got one reservation about this which is that the site name is included in the strings for translations, and the pull request of @dracos [here](https://github.com/mysociety/yournextmp-popit/pull/402) will mean that viewing YourNextMP with `Accept-Language: es` will mean that you'll get "YoQuieroSaber" instead of "YourNextMP". Perhaps it should either be something in `elections/*/settings.py` or just use the name from the sites framework. On the other hand, there's probably quite a bit of election specific stuff in the translations still, so it's part of a larger issue.

Anyway, I think it's best that you go ahead and squash and merge this, and we can deal with that later.

:+1:

PR #392, opened 2015-06-12
Replace more UK-specific code

mhl:
This pull request includes a variety of fixes to make the code increasingly
less UK-specific; in particular:

* The constituency (or "candidates for a particular post") view is no longer
  UK specific.

* Many uses of the MapIt ID where it should be a post ID in the generic
  case have been fixed.

* When editing a person the post (constituency) drop-down and party
  drop-downs should only be shown when the person is marked as standing.
  Also, the party drop-down that's shown should be in the right "party
  set" for the post.  This has been broken since we changed this page so
  that candidacies for multiple posts could be shown and edited.

* In addition, this adds a simple party set for Argentina, where all
  parties are in the 'nacionale' party set.

This pull request includes party JSON for Argentina from Martín.


candidates/static_data.py
```diff
@@ -63,12 +63,19 @@ class MapItData(object):
...
-        areas_by_id[t] = get_mapit_areas(t[0], t[1])
+    for t, election_data in MAPIT_TYPES_GENERATIONS_ELECTIONS.items():
+        mapit_type, mapit_generation = t
+        areas_by_id[t] = get_mapit_areas(mapit_type, mapit_generation)
         for area in areas_by_id[t].values():
             areas_by_name.setdefault(t, {})
             areas_by_name[t][area['name']] = area
+            post_id = election_data['get_post_id'](mapit_type, area['id'])
+            if post_id in areas_by_post_id:
+                message = "Found multiple areas for the post ID {0}"
```
dracos: Could we try and i18nize all strings from now on?

mhl: Thanks, I've pushed a fixup to do that

candidates/static_data.py
```diff
@@ -106,7 +114,3 @@ def get_all_parties():
 
 class PartyData(object):
     party_choices, party_id_to_name = get_all_parties()
-    party_sets = (
```
dracos: get_all_parties  still hardcodes GB/NI, so I'm not sure 'nacionale' exists anywhere to look up?

dracos:
Looks good, some minor comments :)

PR #394, opened 2015-06-16
Add ElectionMixin for URLs using an election slug.

### LINKED ISSUE #380: consider reusing code from ElectionListView for other views
Either as a mixin or a context processor, Matthew pointed out that this could be reused anywhere wherer `election` is in `kwargs`:

https://github.com/mysociety/yournextmp-popit/commit/0c042a802b7bc0bf7be107e2f2f4349164912c4e#commitcomment-11631220

dracos:
Fixes #380.

### DIFF
--- a/candidates/views/candidacies.py
+++ b/candidates/views/candidacies.py
@@ -6,6 +6,7 @@ from braces.views import LoginRequiredMixin
 from auth_helpers.views import user_in_group
 
 from candidates.models import get_area_from_post_id
+from elections.mixins import ElectionMixin
 
 from .helpers import get_redirect_from_mapit_id
 from .version_data import get_client_ip, get_change_metadata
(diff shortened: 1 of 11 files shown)


candidates/views/candidacies.py
```diff
@@ -51,19 +52,19 @@ def form_valid(self, form):
         )
         # Update standing_in and party_memberships:
         new_standing_in = person.standing_in.copy()
-        new_standing_in[election] = get_area_from_post_id(
+        new_standing_in[self.election] = get_area_from_post_id(
```
mhl: I don't think `self.election` will be set here with the code as it is, but it would be nice if `ElectionMixin` also overrode `dispatch` to set `self.election`, `self.election_data` and had the 404 check there instead of in `get_context_data`.

mhl: Nonsense, that's exactly what it does already - I was looking at the code that was being taken out, sigh.

mhl:
Looks good, thanks! :+1:

Issue #395, opened 2015-06-17
Styling for Argentina site


zarino:
Make a theme for Argentina based on the YNMP theme, but with Congreso Interactivo’s colours.

It's likely that we'll have more themes like this in future. So, it'd be a good idea to make sure raw colour variables (eg: `$color-red`) aren't used in any of the core CSS declarations, and instead use generic variable names (eg: `background-color: $header-background-color;`) which cobrands can easily override.

Once this is done, it should be trivial to create a new theme which just inherits from the core theme but overrides the default colours.

zarino:
Pull request with my work so far: https://github.com/mysociety/yournextmp-popit/pull/400

mhl:
Just to add a small to-do, `icon-candidates.gif` doesn't work well as it stands due to having a white background:

![icon-candidates](https://cloud.githubusercontent.com/assets/7907/8253829/4d6b983a-168b-11e5-8aa8-af2c0af27e9e.png)

JenMysoc:
Ok having spoken to Martin homepage needs to include;

Space for text about goal of the research and call to action to get data put in
Form to enter postcode or location (or to auto locate) - same as the UK one
Space for presentation - probably in powerpoint or google sheets - showing volunteers how to participate (This will be available from July 3rd 2015 - not currently finished)

The footer needs to have some space for 4 organisation logos and a list of collaborating orgs (to be provided)

zarino:
@JenMysoc @martinszy is roughly what you're thinking of?

![homepage 1](https://cloud.githubusercontent.com/assets/739624/8430135/32e763ec-1f26-11e5-90ca-32c3b300ca53.png)

martinszy:
That looks great.

zarino:
@martinszy I've just pushed a change to [branch issues/395-homepage-styling](https://github.com/mysociety/yournextmp-popit/tree/issues/395-homepage-styling) that implements a layout like the wireframe above.

![screen shot 2015-06-30 at 16 52 34](https://cloud.githubusercontent.com/assets/739624/8435351/b6c52b8c-1f48-11e5-83ed-861c7fc0798b.png)

Let me know what you think.

martinszy:
I like it, the "why not start with" should diplay the 5 posts that have less candidates.
I'm worried about the form with location search, I hope it works for any address.
We should also add a link to a list of provinces there ... or maybe the posts page.

martinszy:
Also, a few ideas about the header:
[...]

PR #400, opened 2015-06-18
Beginnings of a theme for Yo Quiero Saber

zarino:
Not sure whether we want to merge this as-is, or what until the theme is "complete".

Anyway, this is part of #395, and creates a sub-theme for YQS, which mostly inherits from the core YourNextMP styles, but overriding the colours.

screen shot 2015-06-18 at 15 29 38


martinszy:
Hi, a comment on the menu: Maybe sign in/out options should't have the same hierarchy as the rest of the menu. These could be on the top right corner, as this location is becoming common for them among many websites.
Also, our logo will have a different slogan for this site. The "Yo quiero saber / elecciones 2015" is for the voter information site, and this site will be "Yo quiero saber / investigación". We can either do that by changing the image or by having a text slogan in the html.

zarino:
Hey @martinszy! Thanks for the feedback. Don't worry about the menu – I expect this will be going through a few iterations before it even comes close to being finished. The aim of this pull request is just to get a *really basic* theme in place, with as *simple* markup and styles as possible, so we can build on it later. That includes thinking about stuff like what should go in the header, what the information architecture of the site is going to be, and how prominent different menu items should be.

Swapping the image out will be easy. Do you have the correct one for me to use right now?

I'd suggest @mhl gets this merged into master now, and we can open new branches and pull requests for subsequent changes as required.

PR #402, opened 2015-06-19
Restrict LANGUAGES to those we have.

dracos:
Without this, as our two current languages are en and es-ar, if someone
presents an Accept-Language: es header they get en (because es is a
supported Django language).

Plus one PPC rename.

### DIFF
--- a/candidates/models/popit.py
+++ b/candidates/models/popit.py
@@ -117,7 +117,7 @@ form_complex_fields_locations = {
         'sub_array': 'links',
         'info_type_key': 'note',
         'info_value_key': 'url',
-        'info_type': 'party PPC page',
+        'info_type': 'party candidate page',
     }
 }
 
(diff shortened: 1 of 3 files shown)


mhl:
> Without this, as our two current languages are en and es-ar, if someone
presents an Accept-Language: es header they get en (because es is a
supported Django language).

I'm probably missing something, but with 453b8e12f54c816f4d (i.e. without this PR) I see the Spanish strings when presenting an `Accept-Language: es,en-US;q=0.7;q=0.3` header:

language-es

Is that because I have `LANGUAGE_CODE: 'es-ar'` in my config? If so, it seems like this is working fine anyway, since anyone who's working on the code presumably will have that setting in their `general-ar_elections_2015.yml` file.

dracos:
"Is that because I have LANGUAGE_CODE: 'es-ar' in my config?" <-- yes, presumably, this PR is independent of the LANGUAGE_CODE setting. If someone visits YourNextMP with their browser configured for Spanish, why not give most of it in Spanish if we're able to? :)

mhl:
I've cherry-picked 6410585 as 4933b148b1, and added d2c58d926bb484164

mhl:
I've created [a pull request with an alternative to the first commit](https://github.com/mysociety/yournextrepresentative/pull/436), so I'm going to close this PR - hope that's OK.

PR #409, opened 2015-06-30
Update primary candidates representations and more work to remove UK-specific code

mhl:


chrismytton:
Looks good to me :+1:

PR #410, opened 2015-06-30
More tests for the /areas and /posts pages

### LINKED ISSUE #379: add more tests for the /areas and /posts pages
There are just smoke tests for this at the moment; they just check that the pages load without error, not checking anything about the page content.

lizconlan:
Still a bit smoky in places but, particularly for areas, pokes the underlying logic and checks on what the page should be dealing with a bit more.

Includes a couple of template and view fixups which were caught by the new tests

Closes #379


mhl:
:+1: Looks good, nice and thorough tests

PR #411, opened 2015-06-30
Fix the add candidate button

### LINKED ISSUE #388: fix the add new candidate buttons in candidate lists when you're not logged in
These still say "add new candidate" when they should probably be "log in to add a candidate" if the user isn't logged in.

### LINKED ISSUE #408: Add a new candidate button doesn't work
The button to add a new candidate does nothing currently

http://argentina.ynr.staging.mysociety.org/election/senadores-argentina-paso-2015/post/sen-prim-3/candidato-a-senador-nacional-por-catamarca as an example.

lizconlan:
This seemed to just be a case of adding a second version of the button which would only be displayed if the user isn't allowed to edit (assume not logged in). Could just as easily become an only show button if statement.

Includes updated .po files which I had expected to just add "Sign in to add a new candidate" and yet... Hmmm. Will take that under advisement

Closes #388
Closes #408

### DIFF
--- a/candidates/example-popit-data/posts_65730_embed=membership.person.membership.organization.json
+++ b/candidates/example-popit-data/posts_65730_embed=membership.person.membership.organization.json
@@ -1,0 +1,19 @@
+{
+  "result": {
+    "area": {
+      "id": "mapit:65730",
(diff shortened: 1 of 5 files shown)


mhl:
:+1: Looks good, thanks!

PR #412, opened 2015-07-01
Add a page showing the posts with most attention needed

### LINKED ISSUE #404: Create page for YNR Argentina showing elections in order of data priority
From Martín S: We will need a page that has the all the elections ordered in a way that shows the priority of those which need to be worked on, first. A table of cities (and provinces) that shows the amount of candidates entered for each of them, those with no candidates will be on top for new users to take. This will allow the users to set their own priorities without asking us. Mark has shown me there is something very similar already, we would just need to have a text before to explain this in the page for volunteers.

mhl:
These are the posts ordered by the number of candidates, from
lowest to highest; this is a reasonable proxy in many cases for
those that need most attention from contributors.
    
This has been particularly requested for Argentina.

This is part of the work to fix #404 (the other bit is to include the
first few items from this list on the front page).

### DIFF
--- a/cached_counts/templates/reports.html
+++ b/cached_counts/templates/reports.html
@@ -9,6 +9,16 @@
 
 {% block content %}
 
+  <h2>{% trans 'Which posts have fewest candidates so far?' %}</h2>
+
+  <p>{% url 'attention_needed' as url %}
+     {% blocktrans %}
+        You can see a list of all the posts in current elections
+        <a href="{{ url }}">ordered starting with those with the fewest
+        candidates</a>.
+     {% endblocktrans %}
+  </p>
+
   {% for era, elections in all_elections.items %}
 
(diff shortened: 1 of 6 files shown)


cached_counts/templates/reports.html
```diff
@@ -9,6 +9,16 @@
 
 {% block content %}
 
+  <h2>{% trans 'Which posts have fewest candidates so far?' %}</h2>
+
+  <p>{% url 'attention_needed' as url %}
+     {% blocktrans %}
+        You can see all posts the posts in current elections
```
chrismytton: "posts the posts" doesn't seem quite right here, is the first "posts" not needed?

chrismytton:
Code looks fine, just a small typo :wink: Otherwise looks good :+1:

mhl:
Oops, thanks @chrismytton

PR #422, opened 2015-07-02
Update API help page to generate a link for each csv file

### LINKED ISSUE #416: update the API help page to refer to the new "file-per-election" and "all election" CSV files
Since https://github.com/mysociety/yournextmp-popit/commit/3054a83525610e4ffd544f877b727b7309544195 the name of the generated CSV file(s) will be different - it used to just generate one `candidates.csv` file. Now it'll generate one CSV file per election (with the election slug from the `ELECTIONS` setting in the filename) and one file with all elections' data in it, e.g.:

* `candidates-all.csv`
* `candidates-2010.csv`
* `candidates-2015.csv`

However, the [API / Download page](https://github.com/mysociety/yournextmp-popit/blob/master/candidates/templates/candidates/api.html) just links to the old `candidates.csv` file. It would be good to update that so that it links to a CSV file for each of the current elections, then the older elections and then the `-all.csv` file.

lizconlan:
Previously here was only ever a single candidates.csv file. Now that we're supporting having more than one election at a time, that's no longer the case.

Fixes #416

### DIFF
--- a/candidates/templates/candidates/api.html
+++ b/candidates/templates/candidates/api.html
@@ -23,14 +23,40 @@ out more about PopIt here</a>.)
 
 <h2>{% trans "CSV/Excel Download" %}</h2>
 
+<h3>{% trans "Current elections" %}</h3>
+
 <p>
   <ul>
-    <li><a href="{{ MEDIA_URL }}candidates-all.csv">{% trans "Download of the candidates for all elections" %}</a></li>
-    {% for csv in csv_list %}
+    {% for csv in current_csv_list %}
       {% with slug=csv.slug title=csv.name %}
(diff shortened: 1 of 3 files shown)


mhl:
Just one minor comment - the per-election links are only for current elections, but the `-all.csv` download contains historic elections too. I think it might be clearer to lay it out as:

## CSV/Excel Download

### Current elections

* Download of the Diputados Nacionales PASO 2015 candidates
* Download of the Gobernador PASO 2015 candidates
* Download of the Presidentes PASO 2015 candidates
* Download of the Senadores Nacionales PASO 2015 candidates

### Historic elections

* Download of the ....

### All elections

* Download of the candidates for all elections (current and historic)

lizconlan:
Ah, yes - good point, thanks!

lizconlan:
( these should probably be squashed into a single commit )

mhl:
:+1: Looks fine - thanks!

PR #430, opened 2015-07-03
Remove underline on buttons on hover.

### LINKED ISSUE #425: AR: text inside a button should not be underlined on hover

zarino:
Fixes #425.


mhl:
:+1: Great, thank-you

PR #432, opened 2015-07-06
Add fourth footer logo

zarino:
Adds the Directorio Legislativo to the footer. Makes all the other logos a bit smaller so it can fit alongside.


mhl:
:+1: (Although to be pedantic, it'd be nice to have the comment about also making the other logos smaller in the commit message too, not just the PR message :))

Issue #433, opened 2015-07-06
Yo Quiero Saber is missing a logo suitable for twitter/facebook cards


zarino:
The YourNextMP one was 500x500 pixels.

PR #434, opened 2015-07-06
Add Yo Quiero Saber logo for facebook/twitter cards

zarino:
For #433.


mhl:
One small thing to change, otherwise :+1:

mhl:
Just for the record, this was inattentive reviewing on my part - it also needed the `{% load metadescription %}` and the path of the file should have been `elections/ar_elections_2015/static/img/logo.png`; I've rewritten the commit on master to change those, since we were rebasing and force-pushing anyway.

PR #435, opened 2015-07-06
Update the posts to be invalidated any time standing_in is changed

mhl:
The cache of post (with embedded memberships, people and organizations)
wasn't being invalidated properly on creating new people, in particular
when a new person was created or had their membership of posts changed.


lizconlan:
:+1: looks good :)

PR #436, opened 2015-07-08
Remove 'PPC' from the link key for a party candidate page

mhl:
Thanks to Matthew Somerville for spotting this still used UK-specific
terminology.


dracos:
:+1:

Issue #353, opened 2015-04-28
mark up strings on the site for translation


mhl:
See:
- https://docs.djangoproject.com/en/1.7/topics/i18n/translation/

Obviously that won't be enough to localize the site for Argentina, but it's a necessary pre-requisite. Some care will be needed to make sure that format strings make sense in isolation e.g. `"Information about {full_name}"` rather than `"Information about {0}"`

martinszy:
I started this, but I can't get the site to take the strings from the po/mo files.
https://github.com/YoQuieroSaber/yournextmp-popit

I've also created a document in the wiki commenting on the process I followed: 
https://github.com/YoQuieroSaber/yournextmp-popit/wiki/Translation-process

dracos:
I have successfully viewed the site with your Spanish translations :) I have uploaded the translation to Transifex which you can find at https://www.transifex.com/projects/p/yournextmp/ , in case that's a preferable way of translating. Or you can edit the .po file in locale/es_AR directly as you have done in your fork. I think one issue you had was that you want a LANGUAGE_CODE of `es-ar`, not `es_AR`. I have added code to our branch to allow you to set a default fallback language/timezone in the `general.yml` file rather than settings.

As you say in your wiki, you can run `compilemessages` when you add translations; the other steps should all have been done for you now on our master branch.

JenMysoc:
A string on the Email confirmation page is untranslatable still:

"We have sent an e-mail to you for verification. Follow the link provided to finalize the signup process. Please contact us if you do not receive it within a few minutes."

dracos:
That's in a third party library (allauth) where the particular translation has been marked as fuzzy. That is fixed in the 0.21 release (we are currently using 0.18), so if an upgrade is possible, that should fix it.

mhl:
@dracos I've updated allauth in [this pull request](https://github.com/mysociety/yournextrepresentative/issues/446) and put it on the staging site, but this message is still coming up in English, even with `es` first in my `Accept-Languages`:

* http://argentina.ynr.staging.mysociety.org/accounts/confirm-email/

Or testing on the command line with:

    curl -H 'Accept-Language: es' 'http://argentina.ynr.staging.mysociety.org/accounts/confirm-email/'

Is there a step I'm missing? (`./manage.py compilemessages` has been run as part of the deploy.)

dracos:
[...]

PR #446, opened 2015-07-17
Package upgrades: Django to 1.8.3, allauth to 0.21.0 and others

mhl:
The motivations for this upgrade are two-fold:

* We need a more recent version of allauth to get translatable
  localizable versions of its text, as pointed out by Matthew
  here:

    https://github.com/mysociety/yournextrepresentative/issues/353#issuecomment-122213233

* We should be on Django 1.8 since it's a Long-Term Support
  release and 1.7 will stop getting security updates at some point
  soon.

These upgrades also requires some other upgrades either as direct
dependencies or in order to get rid of new warnings and errors,
in the case of:

  * django-debug-toolbar
  * django-nose

We also needed to remove South, which is no longer used, and was
generating warnings.

Automatically migrating to this version will fail, since allauth now has
Django migrations rather than South migrations and it doesn't know that
various relations have already been created.  You will need to fake two
of these migrations with:

  ./manage.py migrate --fake account 0001
  ./manage.py migrate --fake socialaccount 0001_initial

In order to suppress a deprecation warning about supplying both
'auto_now_add' / 'auto_now' and 'default' for DateTimeField fields
on QueuedImage, this also includes a migration to remove the 'default'.


dracos:
You can remove `SOUTH_TESTS_MIGRATE` from settings I assume. Looks good :)

PR #447, opened 2015-07-17
Issues/395 argentina style fixes

zarino:
Connected to #395 – some small CSS fixes for the Yo Quiero Saber header and homepage.

Also, adds a minor fix to the core YNR styles, to cap the width of the site at something sensible in IE7–8 (where the current `max-width`, set in `rem`s, doesn't work).


mhl:
:+1: Thanks, that's merged now

PR #467, opened 2015-07-22
Make per-province "party sets" work for Argentina

mhl:
This introduces a "party set" per province; these party sets have
all the parties in the national party set ('nacional') as well as
some specific to that province.

This means that the right party list should be shown when editing
a candidate - if you change the post that they're standing for then
Javascript will show a different party list specific to that other
province.


dracos:
Other than a typo in one commit message, looks good :)

PR #476, opened 2015-07-27
Try and always use standing_in property.

### LINKED ISSUE #475: Adding new candidate sometimes creates bad data, or gives error
Steps to reproduce:
1. http://investigacion.yoquierosaber.org/areas/PRV-15/buenos-aires, "Agregar un candidato nuevo" to e.g. "Diputado Nacional por BUENOS AIRES".
2. Set name and source to TESTING, leave party blank, click Agregar.
3. On error page (as you didn't give a party), set standing to:
    a. No and submit to get an immediate 500 on the submission
    b. Don't know and submit to get it to create an entry with bad data, such that trying to load the new person (after the "success" redirect) now gives a 500 error.

dracos:
It makes sure to always return a dictionary no matter what might be
stored. Also check we have some standing_in contents before testing a
lock on it.
Might fix #475

### DIFF
--- a/candidates/models/auth.py
+++ b/candidates/models/auth.py
@@ -47,7 +47,7 @@ def get_constituency_lock(user, api, post_id):
 def check_creation_allowed(user, api, new_popit_data):
     for election in settings.ELECTIONS:
         standing_in = new_popit_data['standing_in']
-        if election not in standing_in:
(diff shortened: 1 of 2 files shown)


mhl:
:+1: Thank-you

PR #477, opened 2015-07-28
Some changes for argentina production site

martinszy:


mhl:
Hi @martinszy - I've made a few small suggestions on those commits, please let me know if any of that is unclear - otherwise it all looks great.

PR #479, opened 2015-07-29
YoQuieroSaber needs old GA for the moment; add a config switch for that

mhl:


jacksonj04:
😧

jacksonj04:
My only comment would be that given Universal is the new standard, it might be more sensible to call it `USE_LEGACY_ANALYTICS` and invert the logic (ie you explicitly say you want the legacy version, rather than the current standard).

Other than that, :+1:

PR #480, opened 2015-07-29
Fix loss of alpha channel by converting cropped images to RGBA not RGB

### LINKED ISSUE #478: processing of images in the moderation queue breaks transparency
For example, we've uploaded the following image of Manuela Castañeira from:

* http://elecciones2015.yoquierosaber.org/static/img/candidatos/manuela-castaneira.png

Original image of Manuela Castañeira

... but once approved, this image is broken on the site:

broken-image-transparency

Some quick testing suggests that this is due to the processing with Pillow here:

* https://github.com/mysociety/yournextrepresentative/blob/509e570acf73e1c38d48e43a2000c3ab57c5aed1/moderation_queue/views.py#L210-L216

... rather than the PopIt image proxy, since the original image is can be proxy without those artefacts:

* https://ynr-argentina.popit.mysociety.org/image-proxy/http%3A%2F%2Felecciones2015.yoquierosaber.org%2Fstatic%2Fimg%2Fcandidatos%2Fmanuela-castaneira.png/100/100

fine through proxy

mhl:
Fixes #478

### DIFF
--- a/moderation_queue/views.py
+++ b/moderation_queue/views.py
@@ -209,8 +209,10 @@ class PhotoReview(GroupRequiredMixin, PopItApiMixin, TemplateView):
     def crop_and_upload_image_to_popit(self, image_filename, crop_bounds, moderator_why_allowed, make_primary):
         original = Image.open(image_filename)
         # Some uploaded images are CYMK, which gives you an error when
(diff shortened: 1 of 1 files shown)


lizconlan:
yay! :+1:

PR #481, opened 2015-07-29
Simplify gem installation

### LINKED ISSUE #469: Deploying reinstalls compass each time
It probably doesn't need to do that.

mhl:
Only install sass (compass isn't needed) and only install it if the
required version isn't already present.

Fixes #469


lizconlan:
one very minor comment, :+1: otherwise (just tested it on local VM)

PR #488, opened 2015-07-31
WiP: Canonicalize BASE_DIR, which should fix some caching issues

mhl:
The keys used in the memcached backend are based on the database name
(so that multiple instances of the application on the same host can
use the same memcached); however, in the case where you're using
sqlite as a database, the name of the database will be the path to
the sqlite database file.  This was based on BASE_DIR in
mysite.settings, which didn't canonicalize the path, so sometimes
the path to the database would be relative and sometimes absolute.
This meant that values stored in memcached couldn't be reliably
retrieved.
    
We think this is probably the reason that collecstatic would sometimes
appear not to update the cache of fingerprinted filenames, and may
also explain why cache invalidation for posts seems to be failing on
the instance for Argentina.
    
This commit introduces a change so that BASE_DIR is set to a value
that has has os.path.realpath run on it, so that you always get the
canonical absolute path, which the sqlite database path is then based
on.

### DIFF
--- a/mysite/settings.py
+++ b/mysite/settings.py
@@ -13,11 +13,11 @@ from django.conf.global_settings import TEMPLATE_CONTEXT_PROCESSORS, LANGUAGES
 from django.utils.translation import to_locale, ugettext_lazy as _
 from collections import defaultdict
 import importlib
-import os
+from os.path import dirname, exists, join
 import re
 import sys
 import yaml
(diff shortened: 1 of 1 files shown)


dracos:
Double "has" in commit message, but otherwise great, well done for spotting, hope it is the cause! :)

PR #513, opened 2015-09-11
Make the format string for post labels localizable and overridable

### LINKED ISSUE #510: Create localizable string for PopIt Post labels
This: https://github.com/mysociety/yournextrepresentative/blob/master/candidates/management/commands/candidates_create_popit_posts.py#L26 seems to have been hard coded at some point to assume Spanish. Let's make that localizable.

mhl:
Fixes #510


lizconlan:
:+1: looks good to me!

Issue #474, opened 2015-07-27
Fields cut off if they're too long


dracos:
The field with data should read "Candidato actualmente en Diputados Nacionales PASO 2015" (there should probably only be one if we have it for this set (that is #472), but it's possible there'd be more than one and anyway it still shouldn't cut it off if there's only one).
http://investigacion.yoquierosaber.org/person/46/emilio-martinez-garbino
![screen shot 2015-07-27 at 13 44 45](https://cloud.githubusercontent.com/assets/154364/8906333/b51052da-3465-11e5-9030-605bce230f97.png)
![screen shot 2015-07-27 at 13 46 03](https://cloud.githubusercontent.com/assets/154364/8906347/d7fad3a6-3465-11e5-9bfa-3fe4dec0385a.png)

zarino:
Pull request here: https://github.com/mysociety/yournextrepresentative/pull/518

martinszy:
I verified this is fixed in my updated local dev environment.

PR #518, opened 2015-09-17
Add opt-in "key-value-list" style to avoid definition list truncation

### LINKED ISSUE #517: "Contesting" and "Contested" election lines are truncated
You can't see the name of the election on the person view page (e.g. http://investigacion.yoquierosaber.org/person/47 ) - here's an example from Argentina:

clipped

and one from Burkina Faso:

truncated

I discussed this a bit with @zarino - I think my preferred solution out of those we discussed would be to keep the side-by-side responsive definition list for the other sections, but for "Constituencies" (a badly named section now) it would be nice if the `<dt>` came on a full width line before the `<dd>`, also filling the full width (as you get when the small screen width version is triggered)

zarino:
Fixes #517 #474.

screen shot 2015-09-17 at 12 53 52

### DIFF
--- a/candidates/management/commands/candidates_delete_everything_from_popit.py
+++ b/candidates/management/commands/candidates_delete_everything_from_popit.py
@@ -1,0 +1,38 @@
+import json
+
+from candidates.models import PopItPerson
+from candidates.popit import PopItApiMixin, get_base_url
+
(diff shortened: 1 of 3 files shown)


mhl:
:+1: Thanks, @zarino!

PR #523, opened 2015-09-21
Add support for Burkina Faso (including party lists)

mhl:
This also includes moving the site name from translated strings to
the sites framework, and various other small fixes.


struan:
Few small stylistic things but otherwise I think this looks good.

PR #527, opened 2015-09-24
Show only top N members in party lists on post page

### LINKED ISSUE #524: only show the top default_party_list_members_to_show members in party lists
Although most work for party lists was done in https://github.com/mysociety/yournextrepresentative/commit/46b7c4e79f7fecd9 (including adding the `default_party_list_members_to_show members` election settings for Burkina Faso) there's actually no code support for only showing that number of people on the post page.

I'd probably do this by passing that value into `group_people_by_party`, and truncating the list of people in that function - that value should also be set in each party dictionary that's returned, as well as a boolean indicating if it was actually truncated, since that information will be needed when changing the template.

struan:
If party lists are configured we may only want to show the top N candidates in the list on the page for a post. This adds a max_people param to the group_people_by_party function which, if not None, will limit party lists to that number of people, order by party list position.

Fixes #524


struan:
Closing in favour of #528 which is based against correct branch

PR #528, opened 2015-09-24
Show only top N members in party lists on post page

### LINKED ISSUE #524: only show the top default_party_list_members_to_show members in party lists

struan:
If party lists are configured we may only want to show the top N candidates in the list on the page for a post. This adds a max_people param to the group_people_by_party function which, if not None, will limit party lists to that number of people, order by party list position.

Fixes #524


candidates/views/helpers.py
```diff
@@ -92,20 +92,40 @@ def group_people_by_party(election, people, party_list=True):
...
-        party_id_to_people[party_id].append(person)
-    for people_list in party_id_to_people.values():
-        people_list.sort(key=lambda p: p.last_name)
+        party_id_to_people[party_id].append((position, person))
+    for party_id, people_list in party_id_to_people.items():
+        if election_data['party_lists_in_use']:
+            people_list.sort(key=lambda p: ( p[0] is None, p[0] ))
+        else:
+            people_list.sort(key=lambda p: p[1].last_name)
+        # TODO: should this only be if party_lists_in_use?
```
struan: Currently it truncates the list if it's not using party lists. I suspect it should not.

mhl:
Looks good...

The Travis failures are because 'part_lists_in_use' is missing from the elections apps other than Burkina Faso. (I'm *convinced* that I went through and added these settings for UK and AR, but I can find no evidence of that now.)

As you pointed out, this doesn't include the template changes, so it would be good to add them too.

I'd probably squash all these commits together after that; it feels like no one part is useful independently of the others (although keeping your helpful commit messages in the new commit message!)

candidates/views/helpers.py
```diff
@@ -92,20 +92,41 @@ def group_people_by_party(election, people, party_list=True):
     party."""
 
     party_id_to_people = defaultdict(list)
+    party_truncated = defaultdict(list)
```
mhl: This can just be a `dict` and the reference to `party_truncated[k]` below be changed to `party_truncated.get(k)` so that if it's not present you get `None` back rather than a `KeyError` exception.

PR #533, opened 2015-09-25
Add St. Paul, Minnesota 2015 municipal elections

evz:
This adds a single election to the collection for City Council and School Board Members in St. Paul, Minnesota in November 2015. A major difference between this and past examples is that it uses [Represent](http://represent.poplus.org/) instead of MapIt as the source for looking up geographies.


candidates/static/candidates/_content.scss
```diff
@@ -11,9 +11,9 @@
   padding-right: $column-gutter / 2;
 }
 
-.button, input[type="submit"].button {
-  @extend .radius;
-}
+//.button, input[type="submit"].button {
+//  @extend .radius;
+//}
```
mhl: I think the error that would have suggested commenting this out was caused by using a more recent version of sass than is supported - now that you've switched back to 3.2.19, this scss should be restored. (I've created a ticket to [replace that scss with some that works with newer versions](https://github.com/mysociety/yournextrepresentative/issues/537) - hopefully one of our designers will look at that soon.)

elections/st_paul_municipal_2015/views/redirects.py
```diff
@@ -0,0 +1,60 @@
+from django.views.generic import RedirectView
```
mhl: These redirects are a legacy from changing the structure of URLs on the UK YourNextMP site, so you shouldn't need this file (or the import from `elections/st_paul_municipal_2015/views/__init__.py`)

elections/st_paul_municipal_2015/views/parties.py
```diff
@@ -0,0 +1,16 @@
+from candidates.models import get_identifier
```
mhl: I don't think this should be needed in the `st_paul_municipal_2015` application.

elections/st_paul_municipal_2015/views/constituencies.py
```diff
@@ -0,0 +1,44 @@
+import re
```
mhl: I think this file shouldn't be needed either.

elections/st_paul_municipal_2015/templates/st_paul_municipal_2015/districts.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'candidates/constituency.html' %}
```
mhl: This template (and `elections/st_paul_municipal_2015/templates/uk_general_election_2015/constituency.html`) can be omitted.

elections/st_paul_municipal_2015/templates/base.html
```diff
@@ -0,0 +1,18 @@
...
+  {% if settings.GOOGLE_ANALYTICS_ACCOUNT %}
+    <script>
+      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+      m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+      })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+
+      ga('create', '{{ settings.GOOGLE_ANALYTICS_ACCOUNT }}', 'auto', {'allowLinker': true});
+      ga('require', 'linker');
+      ga('linker:autoLink', ['yournextmp.com'] );
```
mhl: This is actually YourNextMP-specific, so you should remove this block - the template file still needs to exist, though, so that there's a `base.html` for other templates to extend.

elections/st_paul_municipal_2015/mapit.py
```diff
@@ -0,0 +1,55 @@
+# coding=utf-8
```
mhl: This file shouldn't be needed in the `elections/st_paul_municipal_2015` app.

elections/st_paul_municipal_2015/lib.py
```diff
@@ -0,0 +1,64 @@
...
+        return re.sub(r'^Council Member for ', '', post_label)
+
+    def area_to_post_group(self, area_data):
+        return area_data['country_name']
+
+    def post_id_to_party_set(self, post_id):
+        return 'st-paul'
+
+
+EXTRA_CSV_ROW_FIELDS = [
```
mhl: I think from here to the end of the file is UK-specific, so shouldn't be needed.

elections/st_paul_municipal_2015/forms.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
```
mhl: I don't think this file should be needed in this Django application

mhl:
Thank-you very much, @evz, for making this pull request with your work on setting up YNR for St Paul.

As I mentioned in Slack, this PR includes a few commits that are algready in upstream master, so I rebased it onto master and also stripped any trailing whitespace that was introduced - that's our [`datamade-master` branch](https://github.com/mysociety/yournextrepresentative/tree/datamade-master) now. (Apologies for forgetting to mention earlier that we try to avoid any trailing whitespace.) If you're happy with that version, you might want to force-push it to the branch this PR is based on.

The comments I left inline are mostly about files that shouldn't be needed in your version (and can just be removed) which should make the St Paul application smaller, and one about restoring some .scss that had been commented out.

The only other thing I'd suggest is perhaps squashing this down into one commit, since it seems that some of them really represent work-in-progress, and adding the St Paul application feels to me as if it could be considered one atomic change - what do you think?

Otherwise that's all great :+1:

evz:
@mhl I'll go ahead and squash this into one commit and use your datamade-master branch.

evz:
Closing in favor of #538

PR #534, opened 2015-09-25
Fix so that the fingerprinted djangojs.js files can be found

### LINKED ISSUE #525: Photo review image cropping broken on YourNextMP
I don’t know if this is true for YourNextRepresentative in general… But it appears there are javascript errors on the photo review page on YourNextMP that prevent the image cropper from loading:

screen shot 2015-09-21 at 12 55 06

mhl:
The "./manage.py compilejsi18n" command just writes the djangojs.js
files under STATIC_ROOT rather than being processed by
staticfiles, so the fingerprinted versin of this file is never
ceated. It seems that to get staticfiles to process the output of
compilejsi18n you have to follow one of these approaches:

  https://django-statici18n.readthedocs.org/en/latest/faq.html#how-to-configure-static-files-with-django-statici18n

Since we already have mysite/static added to STATICFILES_DIRS this
commit takes the latter approach and gets compilejsi18n to output
to mysite/static/jsi18n/...

We also alter the order of the commands in bin/pre-deploy so that
these files are generated before collectstatic is run.

Fixes #525

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 /conf/httpd.conf
 /conf/httpd.conf.deployed
 /mysite/local_settings.py
+/mysite/static/jsi18n
--- a/bin/pre-deploy
+++ b/bin/pre-deploy
@@ -60,9 +60,11 @@ else
     echo done.
 fi
 
+# Generate the files of translated strings for Javascript to use
+./manage.py compilejsi18n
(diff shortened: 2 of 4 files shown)


dracos:
Typo `versin` in the second commit message, otherwise looks good :)

PR #539, opened 2015-09-30
Avoid Sass compiler error when extending .radius

### LINKED ISSUE #537: change .button extension of .radius so that it works with more recent sass
This is probably a simple thing to fix, but if you use more recent versions of sass than the one that is supported (3.2.19) then you get this error:

```
pipeline.exceptions.CompilerError: Error: ".button" failed to @extend ".radius".
       The selector ".radius" was not found.
       Use "@extend .radius !optional" if the extend should be able to fail.
        on line 14 of /var/www/ynmp/yournextmp-popit/static/candidates/_content.scss
```

People come across this frequently when trying to deploy YNR, so it would be good to change the .scss so that this works on more recent versions too.

zarino:
Fixes #537.

Versions of Sass greater than 3.2.19 would raise an exception (via Django pipeline) when attempting to `@extend` the `.radius` selector from Foundation.

`@include`-ing Foundation’s `radius()` mixin seems to work instead.

---

@mhl – I think this PR fixes the error from the original ticket, but I’m getting pretty broken looking styles in its place. I can't tell whether it's the new line of code (unlikely) or whether it's just a broken dev environment.

screen shot 2015-09-30 at 16 59 20

### DIFF
--- a/bin/pre-deploy
+++ b/bin/pre-deploy
@@ -60,12 +60,12 @@ fi
 mkdir -p "../gems"
 export GEM_HOME="$(cd ../gems && pwd -P)"
 export PATH="$GEM_HOME/bin:$PATH"
-if (sass --version | grep '3\.2\.19') > /dev/null
+if (sass --version | grep '3\.4\.21') > /dev/null
 then
     echo sass is already installed at the correct version
 else
     echo Installing sass...
-    gem install --no-ri --no-rdoc sass -v 3.2.19
+    gem install --no-ri --no-rdoc sass -v 3.4.21
     echo done.
 fi
 
--- a/candidates/static/candidates/_content.scss
+++ b/candidates/static/candidates/_content.scss
@@ -12,7 +12,7 @@
 }
 
 .button, input[type="submit"].button {
-  @extend .radius;
+  @include radius();
 }
 
 .code-sample {
--- a/mysite/settings/conf.py
+++ b/mysite/settings/conf.py
@@ -370,7 +370,7 @@ def get_settings(conf_file_leafname, election_app=None, tests=False):
         'PIPELINE_COMPILERS': (
             'pipeline.compilers.sass.SASSCompiler',
         ),
-        'PIPELINE_SASS_ARGUMENTS': '--trace',
+        'PIPELINE_SASS_ARGUMENTS': '--trace --quiet',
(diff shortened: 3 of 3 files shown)


mhl:
@zarino and I did some investigation of the issue with the CSS being broken this morning.

It seems that when you use a 3.4-series version of Sass with Foundation 3.4, the global Foundation styles aren't included. There are people complaining about that here:

* http://foundation.zurb.com/forum/posts/18898-why-global-styles-are-not-included
* http://foundation.zurb.com/forum/posts/18630-scss-compiles-but-resulting-css-file-incomplete

This works fine with a Sass version from the 3.2 series.

Unfortunately, it seems as if there is no version of Foundation that works with both Sass 3.2 and 3.4 - for example, see the comment for version 5.5 in [the Foundation changelog](http://foundation.zurb.com/docs/changelog.html):

> **Foundation is now compatible with Sass 3.4! Note: this removes Sass 3.2 compatability**

So it looks as if the intent of this ticket (to make it irrelevant whether a user has Sass 3.2 or 3.4 installed) may be an unreasonable expectation. That means the question now is whether we should:

(a) Stick with only supporting Sass 3.2.19 and keeping Foundation at the same version
(b) Upgrade to Sass 3.4.18 and update Foundation to a more recent version that works with that version of Sass. This might mean some upgrade-related bug-fixing and a bit of work for reusers of the code to update their versions, but on the other hand it's probably better for the project to work with the current version of Sass.

wfdd:
You could try using libsass, which would rid you of the sole Ruby dependency.

In requirements.txt, add

```
libsass==0.10.0
```

and in conf.py, replace

```
'PIPELINE_SASS_ARGUMENTS': '--trace',
```

with

```
'PIPELINE_SASS_BINARY': '/usr/bin/env sassc',
```

It works for me, for whatever that's worth.

mhl:
Thanks for the suggestion, @wfdd - I'll test out libsass as a replacement.

(More generally we should deal with this PR soon, it's increasingly awkward to be stuck with these old versions of Sass and Foundation.)

mhl:
I've rebased this onto master, and updated Sass and Foundation to versions that work together. (I've created a new ticket for switching to libsass: #707.

I'd like to merge this really soon, but since it involves a change in depedencies, I'll write an email to known deployers at the same time.

PR #540, opened 2015-10-01
Updates to the call-to-action flash message after updating / creating a person

mhl:
As discussed with @struan / @duncanparkes in IRC, the function to create
this message was less than clear - hopefully switching that to use a partial
template helps with that.

This also fixes a long-standing bug in the rendering, where the flash
message was shown as:

bad-flash

... instead of:

good-flash


struan:
Looks good to me.

Issue #543, opened 2015-10-06
St Paul views need to be updated to use group_people_by_party


mhl:
I think this is the cause of https://gist.github.com/evz/21d828a08f5559d06360 but I'd like to wait for https://github.com/mysociety/yournextrepresentative/issues/542 before working on the fix.

PR #544, opened 2015-10-06
Removing unneeded template

evz:
@mhl I attempted to catch our fork up to your master branch here and squash commits as needed. Let me know if there is something obvious that is missing here.


evz:
@mhl This last commit addresses the error I was getting yesterday. For some reason I thought it was a good idea to override that template, which was entirely unnecessary.

evz:
@mhl I discovered that there was also an issue with our customized area view. The template was expecting the candidates variable in the context to be a tuple of candidates grouped by party and I was still rendering it as a list of candidates.

mhl:
Hi @evz - yeah, I'd created [an issue](https://github.com/mysociety/yournextrepresentative/issues/543) for the grouping-by-party issue, thanks for fixing that.

A lot of the second commit (beb7d86) backs out changes from the first commit (b30671d) - to understand what was being changed I tried squashing the first three commits and then splitting out what seemed like a distinct change to the import candidates script that was also included. I've pushed that rewritten branch to [`possible-datamade-master` in our repository](https://github.com/mysociety/yournextrepresentative/compare/possible-datamade-master?expand=1) which has the same changes but hopefully with a slightly clearer history. If that looks good to you, could you reword the commit messages I marked with FIXME and update this pull request with that branch?

evz:
@mhl I went ahead and swapped out our master for your possible master. The history should be a bit cleaner now.

mhl:
Great, thanks :+1:

Issue #530, opened 2015-09-25
[AR] Priority page. Reduce the relevance of an election according to date.


martinszy:
Elections with a date nearer to the current, should be moved closer to the top.
Sort by date first and by descending amount of candidates within each date-group.

PR #545, opened 2015-10-07
Move elections settings into a model class

struan:
As a first step towards moving to storing the candidate data in the database and not relying on PopIt we need to store election data in the database so we can use that as a key in to the candidates. While we are doing this we might as well move away from storing any election specific data in the settings file and expose it in the Admin interface for easier updating.

This adds an elections model, changes all references to the settings ELECTION data to use the model and removes the ELECTION settings. It also moves the initial election data into fixtures which are loaded in a data migration which is named after the election specific class, using the settings.ELECTION_APP to do so.

This also creates an AreaType model for storing the area types in. This replaces the previous mapit_types array. In order to make the code a bit more readable references to mapit_type etc have been replaced with area_type unless they are specifically referring to MapIt.

Fixes #545


martinszy:
Should we a migration path for existing instalations?
Also, please note we are working on this: https://github.com/mysociety/yournextrepresentative/issues/530 and https://github.com/mysociety/yournextrepresentative/issues/567 which involve modifiying the election settings.

mhl:
@martinszy There's a migration from the existing per-election `settings.py` modules in a6605f5c87c5158dc6466, but that will need to be updated for the new elections you've added and to add the Burkina Faso elections.

mhl:
Nice work, this was a big job.

I've left quite a lot of comments, mostly small things or stuff that might be of interest; the most significant are probably the ArrayField issue and missing elections from the data migration.

Also, before merging I'd like to run this migration and try out the sites for each country we currently support, and I haven't done that yet - since there are a couple of bigger fixups expected, I thought it would make most sense to do that checking after them.

mhl:
That's all the things I spotted when going through that - I think it'd be OK to squash all the commits that are "update * app to use Election model" together, since they have the same kinds of changes in them.

PR #548, opened 2015-10-08
Remove move UK-specific text from the generic applications

mhl:
This removes uses of UK-specific text like 'Democracy Club' in the templates
of the generic application. In addition, this adds generic versions of the
About and Copyright Assignment pages, so that out-of-the-box new deployers
don't get text that refers to the UK general elections, etc.


mysite/settings.py
```diff
@@ -399,6 +399,18 @@
 ELECTION_SETTINGS_MODULE = ELECTION_APP_FULLY_QUALIFIED + '.settings'
 elections_module = importlib.import_module(ELECTION_SETTINGS_MODULE)
 
+# The name of the organisation that runs and manages the site:
+SITE_OWNER = elections_module.SITE_OWNER
```
symroe: Should this have a default?  Or at least be documented somewhere?  Is there a template `ELECTION_SETTINGS_MODULE` somewhere?

symroe:
This looks ace!

I wonder if the DC stuff should be default, or if the default should be a very plain site, with a README that explains why the legal stuff is needed.  We learnt a lot about how and why (in the UK al least, but I expect it's generic enough advice) it's important to Get That Right, so maybe it should be a step that others should be more forced in to thinking about when setting up a new site.

mysite/settings.py
mhl: Not yet, but @davewhiteland suggested that we should create a "simple election" elections app with the minimum settings for people to rename and customize, so I think that's probably how we'll document it. I've created a ticket for that: https://github.com/mysociety/yournextrepresentative/issues/551

In the mean time, I'll add some description to: https://github.com/mysociety/yournextrepresentative/blob/554d692765fa74c0097cb63fc7baaae484f76c61/docs/new-instance.md#define-the-elections-youre-collecting-candidates-for

mhl:
A@symroe Thanks for the review! After this PR the default templates in `candidates/templates` should still have the same copyright assignment and default licensing, etc. but without any of the UK-specific or "history of the copyright confusion debacle" bits, so hopefully new installations will still get the benefit of starting off right unless they customize those templates and remove that.

PR #555, opened 2015-10-12
Import existing bf candidates

### LINKED ISSUE #547: BF: import existing candidate data into Burkina Faso YNR site
It turns out that the team in Burkina Faso already have some candidate data. We'll need to write an import script for this data, as a management command under `elections/bf_elections_2015`. The source of the data is: http://data.gov.bf/organization/ceni-burkina-faso - I had trouble downloading them, but @JenMysoc  emailed me the CSV files in case you're having problems.

The only perhaps non-obvious bit about import scripts is that you shouldn't use the PopIt API directly, since that won't create the versions attribute, etc. - using the PopItPerson model is the best way to do it (e.g. as here: https://github.com/mysociety/yournextrepresentative/blob/master/elections/ar_elections_2015/management/commands/ar_elections_2015_import_candidates.py#L170-L215 )

The site linked to above has new party data, whereas the existing party data in the repository was taken from http://www.ceni.bf/sites/default/files/Legislatives2015/liste_parti_groupement.pdf - it's possible there will be mismatches between them.

Some of the candidates for the presidential post are on Wikipedia, in case that's helpful for cross-checking: https://en.wikipedia.org/wiki/Burkinab%C3%A9_general_election,_2015

struan:
This takes CSVs of existing candidates from the official Burkina Faso site and imports them into YNR. It only imports the candidates and not the 'backup' candidates in the data.

Fixes #547


mhl:
Re-running a slightly earlier version of this script (f60a10f) produced this error:

```
(venv)ynr-bf-staging@pluto:/var/www/burkina-faso.ynr.staging.mysociety.org/yournextrepresentative$ ./m  
anage.py bf_elections_2015_import_candidate  
Traceback (most recent call last):  
  File "./manage.py", line 10, in <module>  
    execute_from_command_line(sys.argv)  
  File "/data/vhost/burkina-faso.ynr.staging.mysociety.org/venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 338, in execute_from_command_line  
    utility.execute()  
  File "/data/vhost/burkina-faso.ynr.staging.mysociety.org/venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 330, in execute  
    self.fetch_command(subcommand).run_from_argv(self.argv)  
  File "/data/vhost/burkina-faso.ynr.staging.mysociety.org/venv/local/lib/python2.7/site-packages/django/core/management/base.py", line 393, in run_from_argv  
    self.execute(*args, **cmd_options)  
  File "/data/vhost/burkina-faso.ynr.staging.mysociety.org/venv/local/lib/python2.7/site-packages/django/core/management/base.py", line 444, in execute  
    output = self.handle(*args, **options)  
  File "/data/vhost/burkina-faso.ynr.staging.mysociety.org/yournextrepresentative/elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py", line 259, in handle  
    'imported_name': party_comp  
  File "/data/vhost/burkina-faso.ynr.staging.mysociety.org/yournextrepresentative/candidates/models/popit.py", line 558, in __setattr__  
    super(PopItPerson, self).__setattr__(name, value)  
  File "/data/vhost/burkina-faso.ynr.staging.mysociety.org/yournextrepresentative/candidates/models/popit.py", line 695, in party_memberships  
    if not is_party_membership(m)  
  File "/data/vhost/burkina-faso.ynr.staging.mysociety.org/yournextrepresentative/candidates/models/popit.py", line 259, in is_party_membership  
    organization_id  
  File "/data/vhost/burkina-faso.ynr.staging.mysociety.org/venv/lib/python2.7/re.py", line 142, in search  
    return _compile(pattern, flags).search(string)  
TypeError: expected string or buffer  
(venv)ynr-bf-staging@pluto:/var/www/burkina-faso.ynr.staging.mysociety.org/yournextrepresentative$ git show -s --oneline
f60a10f BF candidate import management command
```

mhl:
I'm re-running the script but with a line to print out the ID of the person that's being imported to help reproduce this. (It might well be because I killed the first run half-way through - it's possible that would have left `party_memberships` and `memberships` inconsistent.

mhl:
The ID that seems to fail is: `davidkinda-organisationpourladmocratieetleprogrs-1984-12-31`

elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py
```diff
@@ -0,0 +1,274 @@
...
+        mapit_area = mapit_areas_by_name[province]
+        post_id = AREA_POST_DATA.get_post_id(
+            election_id, mapit_area['type'], mapit_area['id']
+        )
+    else:
+        # It must be the presidential election:
+        post_id = ynr_election_data['for_post_role']
+    post_data = get_post_cached(api, post_id)['result']
+    return ynr_election_data, post_data
+
```
mhl: PEP 8 pickiness: there should only be one empty line between function definitions; two empty lines before a class definition. Also I'd remove the doubled empty lines within the bodies of functions.

elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py
```diff
@@ -0,0 +1,274 @@
...
+        mapit_areas_by_name = MAPIT_DATA.areas_by_name[mapit_key]
+        if province != 'Burkina Faso':
+            province = strip_accents(province).upper()
+        mapit_area = mapit_areas_by_name[province]
+        post_id = AREA_POST_DATA.get_post_id(
+            election_id, mapit_area['type'], mapit_area['id']
+        )
+    else:
+        # It must be the presidential election:
+        post_id = ynr_election_data['for_post_role']
```
mhl: I don't this case makes any sense in this version of the script; I'd take out the `else`, the `if province` and remove the default value of `None` for province.

elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py
```diff
@@ -0,0 +1,274 @@
...
+    total = results['total']
+    if total > 1:
+        message = "Multiple matches for party {0}"
+        raise Exception(message.format(party_name))
+    if total == 0:
+        return None
+    # Otherwise there was exactly one result:
+    return results['result'][0]
+
+def unicode_csv_reader(unicode_csv_data, dialect=csv.excel, **kwargs):
```
mhl: I know this example is straight from the docs, but it's really surprising / confusing, I think. `csv.reader` usually takes a file-like object, but this example suggests that all `csv.reader` actually needs it something that will produce a string per row when you iterate over it. Fair enough, I suppose, but it's an assumption about the implementation of `csv.reader`. Also, it's just weird that what happens here is: `str [Windows 1252] --(codecs.open)--> unicode --(encode)--> str [UTF-8]` so that you can passing UTF-8 `str`s to `csv.reader`, and the have to called `decode` on the yielded cell data to turn it back into `unicode`.

So, anyway - I think the only change I'd make here is to add a comment with a link to https://docs.python.org/2/library/csv.html#examples indicating that these two functions are straight from there.

mhl: I'm not necessarily suggesting you make this change, but, for the record, I think what I'd do is something like:

```python
    with open('some-windows-1252-file.csv') as f:
        for encoded_row in csv.reader(f):
            row = [c.decode('windows-1252') for c in encoded_row]
```

(Not tested, but I think that should work.)

elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py
```diff
@@ -0,0 +1,274 @@
...
+
+                    if candidate[fields['birth_date']] is not None:
+                        birth_date = str(dateutil.parser.parse(
+                            candidate[fields['birth_date']], dayfirst=True
+                        ).date())
+
+                    name = first_name + ' ' + last_name
+
+                    id = '-'.join([
+                            re.sub('[^\w]*', '', re.sub(r' ', '-', strip_accents(name.lower()))),
```
mhl: `[^\w]` is the same as `\W`

mhl:
I've added some small comments, but this looks good :+1: and ran through happily on staging.

mhl:
I've just spotted that some people are imported wrong because the party names aren't properly quoted when they contain commas - I think you'll need to fix these lines in listedescandidatsauxelectionslegislativeslisteprovincialeanptic.csv:

line 1355 had had 16 cells
line 1356 had had 16 cells
line 1544 had had 16 cells
line 1545 had had 16 cells
line 1863 had had 16 cells
line 1864 had had 16 cells
line 1865 had had 16 cells
line 1866 had had 16 cells
line 1867 had had 16 cells
line 1868 had had 16 cells

... but adding double quotes around the party name, but one of them also has the comma after the space rather than before. That's meant a few odd candidates like: http://burkina-faso.ynr.staging.mysociety.org/person/4817/bationo-1

candidates/popit.py
```diff
@@ -118,7 +118,7 @@ def get_search_url(collection, query, **kwargs):
     }
     parameters.update(kwargs)
     query_string = '&'.join(
-        k + '=' + urlquote(v) for k, v in parameters.items()
+        "%s=%s" % (k, v) for k, v in parameters.items()
```
mhl: I thought I left a comment about this before, but `%` is on the way out, so it's better style to use `.format`

elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py
```diff
@@ -0,0 +1,274 @@
...
+    help = "Import inital candidate data"
+
+    def handle(self, username=None, **options):
+
+        election_data = {
+                'prv-2015':'listedescandidatsauxelectionslegislativeslisteprovincialeanptic.csv',
+                'nat-2015': 'listedescandidatsauxelectionslegislativesanptic.csv'
+                }
+
+        field_map = {
```
mhl: I think it might be easier to use `csv.DictReader`; it means you can skip the logic to omit the header line as well as these definitions. I think the only two differences are (a) the column name used for region, which could be handled easily and (b) that for one file it's using `NOM_COMPLET` instead of `SIGLE` for the short party name (and since SIGLE is in both files I'm not sure why)


```
`birth_date => DATE_NAISSANCE_TITULAIRE
first_name => PRENOMS_TITULAIRE
gender => SEXE_TITULAIRE
last_name => NOM_TITULAIRE
list_order => ORDDRE_TITULAIRE
party => NOM_COMPLET
party_short => SIGLE
region => PROVINCE_NOM
```

vs

```
birth_date => DATE_NAISSANCE_TITULAIRE
first_name => PRENOMS_TITULAIRE
gender => SEXE_TITULAIRE
last_name => NOM_TITULAIRE
list_order => ORDDRE_TITULAIRE
party => NOM_COMPLET
party_short => NOM_COMPLET
region => CIRCONSCRIPTION
```

elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py
mhl: Yeah, this works fine: https://github.com/mysociety/yournextrepresentative/commit/5c1ce002941

elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py
```diff
@@ -0,0 +1,274 @@
...
+                    birth_date = None
+
+                    if candidate[fields['birth_date']] is not None:
+                        birth_date = str(dateutil.parser.parse(
+                            candidate[fields['birth_date']], dayfirst=True
+                        ).date())
+
+                    name = first_name + ' ' + last_name
+
+                    id = '-'.join([
```
mhl: It doesn't matter in this script, but it's generally not a good idea to shadow the built-in `id` function by creating a variable called `id`.

elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py
struan: So, having made my PEP 8 thing work it seems to want two blank lines between top level functions. I will go and make this all so although good grief, it has a lot of complaints.

elections/bf_elections_2015/management/commands/bf_elections_2015_import_candidate.py
struan: Sadly DictReader and the code to sort out the encoding don't play well together and I suspect making them play nicely together isn't going to be worth the effort.

mhl: DictReader works fine with the simpler decoding code, but you're probably right that it's best just to get this merged and deployed as it is.

PR #557, opened 2015-10-14
Fix errors relating to multiple elections with the same post role

mhl:
We were seeing error emails from YoQuieroSaber that were sent on requests
for pages where the URL specified a post and an election, but that post wasn't
one in that election. These commits fix the underlying issue causing links to
these pages to ever appear on the site and fixes the KeyError exception which
was thrown, which it turns out would be raised in other situations too.


struan:
One small comment aside this looks good.

PR #558, opened 2015-10-16
Posts need an 'elections' array attribute, not an 'election' string

mhl:
The previous change (e9a502a0b) to store an 'election' attribute on
each post actually wasn't quite right, because people can be elected
to the same post in more than one election.  (The example we have at
the moment is that the UK site has both a 2010 and 2015 election
with exactly the same posts.)  The previous change meant that the
UK general election posts only ended up associated with one election
when they should be returned from searches for both.

This commit changes that, so that posts are expected to have a
'elections' array of election IDs instead of an 'election' string
with a single election ID.

After deploying this, you will need to run:

  ./manage.py candidates_create_popit_posts

... to make sure all the posts have the correct 'elections' attribute.

### DIFF
--- a/candidates/example-popit-data/generic_posts_embed=.json
+++ b/candidates/example-popit-data/generic_posts_embed=.json
@@ -11,6 +11,7 @@
       },
       "candidates_locked": false,
       "contact_details": [],
+      "elections": ["2010", "2015"],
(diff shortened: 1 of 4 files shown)


struan:
This looks fine.

PR #561, opened 2015-10-19
Returning validation error when no areas found for St. Paul constituencies

evz:
Just wanted to get this in really quick since it's a somewhat obvious issue


elections/st_paul_municipal_2015/views/frontpage.py
```diff
@@ -62,7 +64,7 @@ def check_address(address_string, country=None):
     except GeocoderError:
         message = _(u"Failed to find a location for '{0}'")
         raise ValidationError(message.format(tidied_address))
-
+
```
mhl: Just one tiny thing - could you remove this trailing whitespace? (No trailing whitespace in source code is [one of the conventions we try to keep to](http://mysociety.github.io/coding-standards.html#indentation-and-whitespace) across the board even if it's not significant in a particular instance; it also means people don't have to change their git config to avoid its obtrusive red highlighting.)

mhl:
Sorry about the delay in dealing with this, @evz  - I've cherry-picked that to master now, where it's: 9c3a9403a88a3295b95bd0d Thanks!

evz:
@mhl Thanks!

PR #562, opened 2015-10-20
Fix translation formatting

mhl:
@evz pointed out that running `./manage.py makemessages` generates lots
lots of spurious changes because it was wrapping lines in the .po files; this
makes it really annoying for contributors to update translations when they
add a new string for localization. This pull request should fix that issue by
standardizing on not word-wrapping either the `msgid` or `msgstr`.


struan:
I've had a look through this and it seems fine.

PR #591, opened 2015-12-17
Update the UK front page's postcode lookup for the 2016 elections

### LINKED ISSUE #583: UK: update postcode lookup to return multiple posts for 2016
At the moment the postcode lookup on the front page goes straight to the post in the 2015 general election - with the 2016 elections, it should redirect to an area page like `/areas/GLA-2247,LAC-11822,WMC-65808` like the generic address lookup in the core code.

mhl:
Fixes #583


struan:
A few minor things but otherwise this looks good.

Issue #586, opened 2015-12-14
add a command to mirror a live YNR instance


mhl:
This would be really useful to help other people get started developing on YNR (and would replace the old `candidates_get_live_database` command).

Essentially, the command would be passed the base URL of a YNR site, and then use the site's API at `/api/v0.1` to fetch all the data from the site and populate it in the local database (assumed to be empty). This will probably need a few extra resources to be added to the API, e.g. the PartySet objects.

PR #592, opened 2015-12-18
Add more models to the API and use more natural IDs

### LINKED ISSUE #585: possibly change the IDs used by the API for organizations, posts, elections and party sets
At the moment, the IDs used by the django-rest-framework-based API for organization, posts and elections are the integer primary keys for the base object, but the IDs used in URLs are actually the (badly named) `slug` fields in OrganizationExtra, PostExtra or Election.

It should be possible to persuade Django Rest Framework to use the extra__slug field as the lookup field (e.g. [people say they got something similar to work](http://stackoverflow.com/questions/29068097/django-rest-framework-lookup-field-through-onetoonefield) although having to define a `user__username` property method as there seems wrong.)

This might make it easier for people to relate results from the API from those from the site. On the other hand, one could say it would make the API implementation more complicated for no real benefit.

(Party sets haven't been added to the API yet, but the same would apply to them.)

mhl:
Fixes #585 and also adds some extra models to help with #586

### DIFF
--- a/candidates/serializers.py
+++ b/candidates/serializers.py
@@ -8,6 +8,22 @@ from images.models import Image
 from elections import models as election_models
 from popolo import models as popolo_models
 
+# These are serializer classes from the Django-REST-framework API
+#
+# For most objects there are two serializers - a full one and a
+# minimal one.  The minimal ones (whose class names begin 'Minimal')
+# are used for serializing the objects when they're just being
+# included as related objects, rather than the resource that
+# information is being requested about.
+#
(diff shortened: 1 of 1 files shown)


struan:
This looks good to me.

Might be nice to have a short comment in the serializers code explaining the minimal serializer pattern used.

mhl:
@struan I've added a comment about the minimal vs full classes 68edf36 and added some fixups so that your tests pass with the changes to the IDs

PR #593, opened 2015-12-18
Upcoming elections api

### LINKED ISSUE #581: add an API which gives upcoming elections (and their areas / posts) given a postcode (or lat/log)
As requested by @symroe - we have all the data to be able to do this on YourNextRepresentative.

Something @tmtmtmtm raised - should this also include a link to a candidates API / YNR instance for each election? (This might be dull for 2016, where it'll all be the DC site, but possibly useful for federation in future and as a prompt that such a thing should always exist :))

struan:
Add an API to list some basic information for the upcoming elections at a postcode or co-ordinate.

The API lists all the positions for which there will be an election in the future, along with details of the area the post is for and the election it is part of. It uses the election_date property of the election to determine which elections to display.

If an election does not have an election_date then it won't be displayed.

Fixes #581

### DIFF
--- a/candidates/mapit.py
+++ b/candidates/mapit.py
@@ -1,8 +1,8 @@
 # coding=utf-8
 
-import re
 import requests
 
+from django.conf import settings
 from django.utils.http import urlquote
 from django.core.cache import cache
 from django.utils.translation import ugettext as _
@@ -45,15 +45,9 @@ def get_known_area_types(mapit_result):
         ]
 
 
-def get_areas_from_postcode(original_postcode):
-    postcode = re.sub(r'(?ms)\s*', '', original_postcode.lower())
-    if re.search(r'[^a-z0-9]', postcode):
-        raise BadPostcodeException(
-            _(u'There were disallowed characters in "{0}"').format(
-                original_postcode
-            )
-        )
-    cached_result = cache.get(postcode)
+def get_areas_from_postcode(postcode):
+    cache_key = 'mapit-postcode:' + postcode
+    cached_result = cache.get(cache_key)
     if cached_result:
         return cached_result
     url = 'http://mapit.mysociety.org/postcode/' + urlquote(postcode)
@@ -61,7 +55,7 @@ def get_areas_from_postcode(original_postcode):
     if r.status_code == 200:
         mapit_result = r.json()
         areas = get_known_area_types(mapit_result)
-        cache.set(url, areas, 60 * 60 * 24)
+        cache.set(cache_key, areas, settings.MAPIT_CACHE_SECONDS)
         return areas
     elif r.status_code == 400:
         mapit_result = r.json()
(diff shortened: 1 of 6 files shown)


candidates/mapit.py
```diff
@@ -1,8 +1,106 @@
...
+        raise UnknownMapitException(
+            _(u'Unknown MapIt error for postcode "{0}"').format(
+                postcode
+            )
+        )
+
+
+def get_areas_from_coords(coords):
+    base_url = 'http://mapit.mysociety.org/'
+    url = base_url + 'point/SRID/' + urlquote(coords)
```
mhl: I think this should be `4326` instead of literally `SRID`

candidates/tests/test_upcoming_elections_api.py
```diff
@@ -0,0 +1,166 @@
...
+            base__organization=london_assembly.base,
+            slug='11822',
+            base__label='Assembly Member for Lambeth and Southwark',
+        )
+        PostExtraFactory.create(
+            elections=(election_gla,),
+            base__area=area_extra_gla.base,
+            base__organization=london_assembly.base,
+            slug='2247',
+            base__label='2016 London Assembly Election (Additional)',
```
mhl: I think the label here should be: `Assembly Member` from https://github.com/mysociety/yournextrepresentative/blob/master/elections/uk_general_election_2015/management/commands/uk_general_election_2015_create_regional_assembly_posts.py#L89-L111

candidates/views/api.py
```diff
@@ -16,6 +17,77 @@
...
+        if errors:
+            return HttpResponse(
+                json.dumps(errors), status=400, content_type='application/json'
+            )
+
+        area_ids = [area[1] for area in areas]
+
+        today = date.today().isoformat()
+        elections = Election.objects.filter(
+            election_date__gte=today
```
mhl: I think the ORM knows about `datetime.date` objects, so you can just do `election_date__gte=date.today()`

candidates/views/api.py
```diff
@@ -16,6 +17,77 @@
...
+            area__identifier__in=area_ids,
+        ).select_related(
+            'area', 'area__extra__type', 'organization'
+        ).prefetch_related(
+            'extra__elections'
+        )
+
+        results = []
+        for post in posts.all():
+            post_elections = [election for election in elections
```
mhl: I *think* it's safe to assume (in practice anyway, even if this isn't enforced in the code) that each post will only be in one current election, so I'd be tempted to change this to `post.extra.elections.get(current=True)` and rather than nesting a list of `elections`, have `election_date` and `election_name` at the same level as `organization`.

candidates/views/api.py
```diff
@@ -16,6 +17,77 @@
...
+                    if election in post.extra.elections.all()]
+            if any(post_elections):
+                results.append({
+                    'name': post.label,
```
mhl: Perhaps `post_name` instead of just `name` here?

mhl:
I've left a few comments that would be worth addressing, but I think this is basically OK - we can deploy it and see what DC / @symroe think.

Other miscellaneous notes:
* I suspect we might want to add more of the election metadata or GSS IDs for areas, but let's get it running and get feedback
* There's now quite a bit of common code between `elections/uk_general_election_2015/mapit.py` and `candidates/mapit.py` - perhaps create an issue to look at that?
* As we discussed in a hangout, this is a third API endpoint which is outside the new `/api/v0.9` namespace - we could make these use DRF's generic views and register them, which I think (I haven't tried this) should make them appear under the same namespace with the same browsable API interface. That's probably something to create a ticket for, though, given that the Christmas break is so soon.

PR #603, opened 2015-12-31
Add a command to mirror a live YNR site from its API

mhl:
This assumes (and checks that you have) empty database tables before you
run the command.

To test this out you'll need to:

* switch country to `example` (because the other defined election apps
   will try to create data when migrating them from scratch)
* drop the database you've defined for the example country
* create a new database of that name
* run `./manage.py migrate`
* run the command with, say:

        ./manage.py candidates_import_from_live_site https://edit.yournextmp.com/

### DIFF
--- a/candidates/management/commands/candidates_import_from_live_site.py
+++ b/candidates/management/commands/candidates_import_from_live_site.py
@@ -1,0 +1,436 @@
+from contextlib import contextmanager
+import errno
+import hashlib
+import json
+from os import makedirs
+from os.path import dirname, exists, join
+import re
+import shutil
+from urlparse import urlsplit, urlunsplit
+
+from PIL import Image as PillowImage
+
+from django.contrib.auth.models import User
+from django.contrib.contenttypes.models import ContentType
+from django.core.files.storage import FileSystemStorage
+from django.core.management.base import BaseCommand, CommandError
+from django.core.management.color import no_style
+from django.db import connection, transaction
+
+import requests
+
+from candidates import models
+from elections import models as emodels
+from popolo import models as pmodels
+from images.models import Image
+
+CACHE_DIRECTORY = join(dirname(__file__), '.download-cache')
+
+PILLOW_FORMAT_EXTENSIONS = {
+    'JPEG': 'jpg',
+    'PNG': 'png',
+    'GIF': 'gif',
+    'BMP': 'bmp',
+}
+
+# n.b. There is some repeated code between here and
+# candidates/migrations/0009_migrate_to_django_popolo.py, but we want
+# to keep the code in the migration frozen, and factoring it out would
+# risk people making changes that broke the migration.
+
+@contextmanager
+def show_data_on_error(variable_name, data):
+    """A context manager to output problematic data on any exception
+
+    If there's an error when importing a particular person, say, it's
+    useful to have in the error output that particular structure that
+    caused problems. If you wrap the code that processes some data
+    structure (a dictionary called 'my_data', say) with this:
+
+        with show_data_on_error('my_data', my_data'):
+            ...
+            process(my_data)
+            ...
+
+    ... then if any exception is thrown in the 'with' block you'll see
+    the data that was being processed when it was thrown."""
+
+    try:
+        yield
+    except:
+        message = 'An exception was thrown while processing {0}:'
+        print message.format(variable_name)
+        print json.dumps(data, indent=4, sort_keys=True)
+        raise
+
+
+class Command(BaseCommand):
+    help = 'Import all data from a live YNR site'
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


candidates/management/commands/candidates_import_from_live_site.py
```diff
@@ -0,0 +1,433 @@
...
+# n.b. There is some repeated code between here and
+# candidates/migrations/0009_migrate_to_django_popolo.py, but we want
+# to keep the code in the migration frozen, and factoring it out qwould
+# risk people making changes that broke the migration.
+
+@contextmanager
+def show_data_on_error(variable_name, data):
+    """A context manager to output problematic data on any exception
+
+    If there's an error when importing a particular person, says, it's
```
struan: says -> say

candidates/management/commands/candidates_import_from_live_site.py
```diff
@@ -0,0 +1,433 @@
...
+        non_empty_models = []
+        for model_class in (
+                # Base Popolo models that YNR uses:
+                pmodels.Person, pmodels.Membership, pmodels.Organization,
+                pmodels.Post, pmodels.ContactDetail, pmodels.OtherName,
+                pmodels.Identifier, pmodels.Link, pmodels.Area,
+                # Additional models:
+                models.PartySet, models.ImageExtra, models.LoggedAction,
+                models.PersonRedirect,
+        ):
```
struan: should this not also check election and area models?

candidates/management/commands/candidates_import_from_live_site.py
```diff
@@ -0,0 +1,433 @@
...
+                            # 'for_post_role',
+                            'current',
+                            'use_for_candidate_suggestions',
+                            'area_generation',
+                            'party_lists_in_use',
+                            'ocd_division',
+                            'description',
+                    )
+                }
+                e = emodels.Election(slug=election_data['id'], **kwargs)
```
struan: What's special about elections where we feel happy replacing/updating things that might already be in the database and not warning if there are existing ones present at the start?

candidates/management/commands/candidates_import_from_live_site.py
```diff
@@ -0,0 +1,433 @@
...
+PILLOW_FORMAT_EXTENSIONS = {
+    'JPEG': 'jpg',
+    'PNG': 'png',
+    'GIF': 'gif',
+    'BMP': 'bmp',
+}
+
+# n.b. There is some repeated code between here and
+# candidates/migrations/0009_migrate_to_django_popolo.py, but we want
+# to keep the code in the migration frozen, and factoring it out qwould
```
struan: typo: 'qwould'

candidates/management/commands/candidates_import_from_live_site.py
```diff
@@ -0,0 +1,433 @@
...
+
+class Command(BaseCommand):
+    help = 'Import all data from a live YNR site'
+
+    def __init__(self, *args, **kwargs):
+        super(Command, self).__init__(*args, **kwargs)
+        self.image_storage = FileSystemStorage()
+
+    def add_arguments(self, parser):
+        parser.add_argument('SITE-URL')
```
struan: A small hint as to what this should be - e.g. "Base URL for live site" might be nice.

candidates/management/commands/candidates_import_from_live_site.py
```diff
@@ -0,0 +1,433 @@
...
+            msg = "This command should only be run on an empty database"
+            raise CommandError(msg)
+
+    def get_api_results(self, endpoint):
+        page = 1
+        while True:
+            url = '{base_url}{endpoint}/?format=json&page={page}'.format(
+                base_url=self.base_api_url, endpoint=endpoint, page=page
+            )
+            self.stdout.write("Fetching " +  url)
```
struan: Two spaces between + and url in case you care :)

candidates/management/commands/candidates_import_from_live_site.py
mhl: I think it's just that I should have added `Election` to the models which we check there are no instances of at the beginning. (Because `Election` instances can be created by migrations (if you have `ELECTION_APP` set to particular values), I initially thought this script should cope with them already existing - but of course, if there is such an `ELECTION_APP` set, the later migration will import data from PopIt and then this script won't work at all.

So I think the simplest thing is to insist there are no elections if you want to use this script.

(As an aside, I'm going to create a ticket suggesting renaming the `uk_general_elections_2015` app to just `uk` - that's a less confusing name, and means that people starting from scratch can use this command to get data, and not be dependent on PopIt.)

struan:
Small queries aside this looks good to me.

Shame about the duplication with the migration but as you say, not much to be done about that :/

Main issue is that on large sites if it falls over it needs to start again which can be quite time consuming.

PR #614, opened 2016-01-06
Import list of Costa Rica parties from CSV file

### LINKED ISSUE #611: costa rica parties import

struan:
Management command and accompanying CSV file for importing a list of parties and generating the related party sets for the 2016 Costa Rica municipal elections.

Takes an argument of a CSV file which should have 3 columns:

    1: Name of party
    2: ID of party (optional)
    3: Comma separated list of Cantons the party is standing in

The ID is the "Cédula Jurídica" and is added as an Identifier.

It expects the CSV file to have NO header row.

It will create, or update, a party for each row in the CSV and a party set for each Canton, adding the party to the party sets for the Cantons in which the party is standing.

This generates slugs by slugifying the party name as we don't have the above IDs for all parties.

If you run

```
/manage.py candidates_create_party_list_from_ep https://cdn.rawgit.com/everypolitician/everypolitician-data/9d6723d/data/Costa_Rica/Assembly/ep-popolo-v1.0.json
```

first then it should match up the imported parties with those pulled from EP.

Fixes #611

### DIFF
--- a/elections/cr/management/commands/cr_import_parties_from_csv.py
+++ b/elections/cr/management/commands/cr_import_parties_from_csv.py
@@ -1,0 +1,135 @@
+# -*- coding: utf-8 -*-
+
+import csv
+import re
+import requests
+
+from django.core.management.base import BaseCommand
+
+from django.utils.text import slugify
+
+from candidates.models import PartySet, OrganizationExtra
+from popolo import models as popolo_models
+from popolo.importers.popit import PopItImporter
+
+
+def fix_whitespace(s):
+    s = s.strip()
+    return re.sub(r'(?ms)\s+', ' ', s)
+
+
+"""
+Takes an argument of a CSV file which should have 3 columns:
+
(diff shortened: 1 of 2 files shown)


mhl:
Sorry, quite a lot of comments!

elections/cr/management/commands/cr_import_parties_from_csv.py
```diff
@@ -0,0 +1,136 @@
...
+
+            OrganizationExtra.objects.create(
+                base=org, slug=slug
+            )
+            print "created new party {0}".format(name)
+
+        if party_id != '':
+            self.add_id(org, party_id)
+
+        for canton in cantons:
```
mhl: `canton` will need to be whitespace-stripped to avoid double spaces in party set names, e.g. `2016 parties in  Abangares Canton`

mhl:
:+1: Great - looks good!

PR #615, opened 2016-01-07
CR: add a script to create posts, elections, area types, etc.

mhl:
(This relies on `PartySet` objects already existing, so should be run after
`cr_import_parties_from_csv.py`.)


struan:
Looks good to me.

PR #616, opened 2016-01-08
CR: use a canton Select2 box Costa Rica front page

mhl:
From discussion with Accesa, it seems that addresses aren't typically
going to useful for geolocation but everyone has the canton printed
on their mandatory government issued ID.

It would be great to do some user testing of this (and also add a
Javascript geolocation option) but for the moment switch the front
page area selection to just have a drop-down list of cantons.


elections/cr/forms.py
```diff
@@ -0,0 +1,99 @@
...
+            ('131', u'Siquirres'),
+            ('132', u'Talamanca'),
+            ('160', u'Tarrazú'),
+            ('161', u'Tibás'),
+            ('116', u'Tilarán'),
+            ('105', u'Turrialba'),
+            ('162', u'Turrubares'),
+            ('96', u'Upala'),
+            ('97', u'Valverde Vega'),
+            ('163', u'Vásquez de Coronado'),
```
struan: Also needs the Zarcero alt name for Alfaro Ruiz.

struan:
Other than missing alt name this looks good to me.

I did notice that Quepos redirects to Aguirre and I think it should be the other way round as Quepos is, I think, the more recent name. However, I assume that's something to fix in the mapit data?

elections/cr/forms.py
mhl: Well spotted - I've pushed a fixup.

mhl:
@struan - Yeah, we should fix it in MapIt and the YNR Area objects. I've created an issue for that, so we can deal with it on Monday: https://github.com/mysociety/yournextrepresentative/issues/619

PR #617, opened 2016-01-08
Tweaks to Costa Rica theme

zarino:


mhl:
Thanks, @zarino - looks good :+1:

PR #626, opened 2016-01-11
Fix a divide-by-zero error in /tasks/{whatever}

mhl:
If there were no candidates (e.g. when the site's first launched)
there would be a divide-by-zero on viewing, say /tasks/email

Thanks to Accesa for reporting this bug.


struan:
Looks good to me.

PR #634, opened 2016-01-11
change candidate id to name on front page recent changes list

### LINKED ISSUE #621: [CR] Removing candidate number from recent changes and replacing with candidate's name
Where the main page lists the Recent Changes, it says, for example: "El usuario xxxxxxx actualizó al candidato #5, hace 1 minuto"

Can this be changed so that the name of the candidate appears, not just his or her number?

struan:
Rather than use the candidate ID on the list of recent changes use the
name of the candidate which makes it a bit more human friendly

Fixes #621


mhl:
Looks good, but could you make the same change to `/recent-changes` at the same time?

PR #635, opened 2016-01-11
Fix non-internationalizable ordinal suffix for day of the month

### LINKED ISSUE #631: [CR] Remove "th" from the end of dates
When you include the day of birth, it appears with a th at the end. Like "11th", this isn't used in spanish

mhl:
The 'S' format specifier for dates is:

  "English ordinal suffix for day of the month, 2 characters."

https://docs.djangoproject.com/es/1.9/ref/templates/builtins/#date

It doesn't make sense having the "th" suffix on a date in Spanish,
for example.

Fixes #631

### DIFF
--- a/candidates/templates/candidates/person-view.html
+++ b/candidates/templates/candidates/person-view.html
@@ -175,7 +175,7 @@
             <small class="dob">({% blocktrans %}Year of birth: {{ dob }}{% endblocktrans %})</small>
         {% else %}
           <dd>{{ person.extra.age }}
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
This seems like something that should be configurable on a per-instance and/or per-locale basis, rather than simply choosing between two hard-coded formats. That would also make it easier to handle countries where dates are usually written in an entirely different manner, e.g. Japanese or Korean style.

mhl:
@tmtmtmtm Fair point - I've pushed a fixup to make it a configurable setting (with a default of `j Fy`) instead, and it's much neater.

struan:
Looks good to me.

mhl:
Oh, grr - actually that new version makes the date format dependent on the site, not the language

PR #636, opened 2016-01-11
If Spanish ('es') is requested for CR or AR, use the site's default

mhl:
It's common for people in Spanish-speaking countries to have
the 'es' language code first in their Accept-Language.  Indeed
Chrome seems not to offer 'es-cr' (Costa Rican Spanish) in its
language preferences at all.

This created a problem because we now have 'es-ar' (Argentinian Spanish)
and 'es-cr' in LANGUAGES; if you request 'es' then whichever was first
in LANGUAGES ('es-ar' with the current code) would always be used.  For
the Costa Rica site, it would make more sense to use 'es-cr' whenever
'es' is requested.

This commit achieves that by looking to see if LANGUAGE_CODE is present
in LANGUAGES, and, if so, moving that entry to the front.  LANGUAGE_CODE
should always be set to 'es-ar' for the Argentinian site and 'es-cr' for
the Costa Rican site, meaning we'll have the desired behaviour.

### DIFF
--- a/mysite/settings/conf.py
+++ b/mysite/settings/conf.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 import importlib
 from os.path import dirname, exists, join, realpath
 import re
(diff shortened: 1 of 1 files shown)


struan:
Looks good to me

PR #638, opened 2016-01-12
Make date-of-birth parsing more liberal

### LINKED ISSUE #165: make the date-of-birth field less picky about date format
For example, we could use dateutil.parser.parse to process the entered date, rather than insisting on ISO 8601

### LINKED ISSUE #444: [optional] Investigate better date format
Could possibly investigate a better date entry format for YQS as the current one has been confusing some of the Argentinian users. 

Perhaps boxes: eg Day [ box to enter day] Month [box to enter month] etc...

mhl:
The change introduced by this commits allows many more date formats
than previously, when just YYYY or YYYY-MM-DD was allowed.

There's also a new DD_MM_DATE_FORMAT_PREFERRED configuration option
to control whether to use the USA or rest-of-the-world date formatting
preference.

Fixes #165, hopefully fixes #444

### DIFF
--- a/candidates/forms.py
+++ b/candidates/forms.py
@@ -12,9 +12,8 @@ from django.contrib.sites.models import Site
 from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 
-from candidates.models import PartySet
+from candidates.models import PartySet, parse_approximate_date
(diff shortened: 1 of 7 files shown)


struan:
This looks good to me.

PR #640, opened 2016-01-12
Add generic support for additional fields

mhl:
We added hard-coded special cases for the 'CV' and 'Program' fields
to add them for Burkina Faso.  This was pretty ugly and not nice to
maintain as we add different additional (non-Popolo) fields for other
countries. For example, Costa Rica want a "Profession" field.

This pull request allows you to add additional fields in the admin
interface which anyone can then edit in the usual way on the site.

Still to do:

- [x] Add a data migration so that if ELECTION_APP is
      bf_elections_2015 then the `cv` and `program` fields
      are added to the database.
- [x] Remove the references to the hard-coded `cv` and
      `program` from the codebase.
- [x] Add a data migration so that if ELECTION_APP is cr
      then the `profession` field is added.
- [x] Update the mirror from live site command to deal with
      additional fields

### DIFF
--- a/candidates/admin.py
+++ b/candidates/admin.py
@@ -3,7 +3,7 @@ from django.contrib import admin
 from django.core.urlresolvers import reverse
 from django.forms import ModelForm
 
-from .models import LoggedAction, PartySet
+from .models import LoggedAction, PartySet, ExtraField, PersonExtraFieldValue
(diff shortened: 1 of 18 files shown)


struan:
This looks good to me.

PR #642, opened 2016-01-12
Remove `key-value-list` class from core profile page list elements

zarino:
The `key-value-list` class stacks a `<dl>` element’s contents vertically, limiting the width of the `<dt>` elements inside and truncating them with an ellipsis if they over-run.

This is mostly ok in English (but not always!) but is a real problem in other languages where important labels get truncated because they’re too long, and there’s no way to reveal the full text.

This commit removes the `key-value-list` class from the `<dl>` elements on the candidate profile page, so the `<dt>` and `<dd>` elements will just stack vertically, as normal.

The CSS rules are left intact, in case re-users want to add the class back into their custom templates.

### DIFF
--- a/candidates/static/candidates/_people.scss
+++ b/candidates/static/candidates/_people.scss
@@ -19,6 +19,8 @@
   border-bottom: 1px solid #ccc;
 }
 
+// You can apply this class to a <dl> element to stack the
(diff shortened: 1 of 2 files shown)


mhl:
:+1: Thanks!

PR #643, opened 2016-01-13
Fix bug where blank birth date was interpreted as today

mhl:
The commit that added dateutil.parser for parsing otherwise
unrecognised dates (507055baf) unfortunately also handled the
empty string and interpreted it as today's date, so not filling
in the date of birth field would give someone today's date as
their birthday.  This commit fixes that and adds regression
tests.


struan:
This looks good to me.

PR #644, opened 2016-01-13
Make sure allauth's .po files are compiled on deploy

mhl:
Because of this bug:

  https://github.com/pennersr/django-allauth/issues/1032

... we need to manually compile the allauth .po files into
.mo files.


struan:
This looks fine to me in as much as I tested the find command worked on my vagrant box.

PR #652, opened 2016-01-14
Fix date parser errors with Spanish

struan:
This should fix handing dates in the translated language and also slightly more fullsome date descriptions like '20 de febrero de 1954'

It does this by adding a parserinfo class with translated month names that is then passed into dateutils.parse to change the set of strings it uses to match months. This uses ugettext_lazy as otherwise the translation is done at module load time which may be before language settings have been changed.


mhl:
A couple of small comments but this looks great, thanks! :+1:

PR #655, opened 2016-01-15
Record / restore extra field values in the versions field

### LINKED ISSUE #647: changes to extra_fields should be recorded to / restored from in the JSON serialization
The file to update is here:

https://github.com/mysociety/yournextrepresentative/blob/master/candidates/models/versions.py

mhl:
The code has been in use for a few days with people being able to
add extra field values without them being recorded in the versions
array, so we should run:

  ./manage.py candidates_record_new_versions

Fixes #647


struan:
Looks good to me.

PR #662, opened 2016-01-18
Add yes/no extra field

struan:
This adds the ability to have extra fields associated with a person that have a dropdown with a Yes/No/Don't know option.

### DIFF
--- a/candidates/templatetags/extra_field_value.py
+++ b/candidates/templatetags/extra_field_value.py
@@ -1,0 +1,28 @@
+from django import template
+from django.utils.html import conditional_escape
+from django.utils.safestring import mark_safe
+from django.utils.translation import ugettext as _
+
+register = template.Library()
+
+
+@register.filter
+def extra_field_value(extra_field):
+    if extra_field['value'] == '':
+        return ''
+
+    if extra_field['type'] == 'url':
+        url = conditional_escape(extra_field['value'])
+        output = \
+            u'<a href="{0}" rel="nofollow">{0}</a>'.format(url)
+        output = mark_safe(output)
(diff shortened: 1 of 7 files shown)


mhl:
I've added a question about the comments and an optional suggestion about factoring out some of the template rendering, but otherwise this looks good :+1:

candidates/templatetags/extra_field_value.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@register.filter
+def extra_field_value(extra_field):
+    if extra_field['value'] == '':
+        return ''
+
+    if extra_field['type'] == 'url':
+        output = \
+            u'<a href="{0}" rel="nofollow">{0}</a>'.format(extra_field['value'])
```
mhl: Unfortunately I think this introduces a cross-site scripting vulnerability, since `extra_field['value']` can be set to whatever a user likes - you could pass it through `django.utils.html.escape` or there's more on that here: https://docs.djangoproject.com/en/1.8/howto/custom-template-tags/#filters-and-auto-escaping

candidates/templatetags/extra_field_value.py
```diff
@@ -0,0 +1,25 @@
...
+    elif extra_field['type'] == 'yesno':
+        if extra_field['value'] == 'yes':
+            output = _(u'Yes')
+        elif extra_field['type'] == 'no':
+            output = _(u'No')
+        else:
+            output = _(u"Don't know")
+    else:
+        output = extra_field['value']
```
mhl: Similarly here

mhl:
:+1: Looks good, thank-you

Issue #605, opened 2016-01-04
add javascript geolocation to YNR


mhl:
It would be nice to have a "show representatives based on my location" button on the front page, particularly for places where reliable addressing and geocoding is available.

zarino:
I assume a "Locate me automatically" type of button under the address/postcode input is what you're after here? Along the lines of FixMyStreet?

struan:
Yes.

PR #665, opened 2016-01-19
Styling for geolocation link on homepage

zarino:
Part of #605. 

Still requires someone to write the actual JavaScript that does the location detection.

screen shot 2016-01-19 at 09 16 13


struan:
Now includes the JS and a view to handle the geolocation.

candidates/static/candidates/_finder.scss
```diff
@@ -181,6 +181,28 @@
...
+  &:before {
+    content: "";
+    display: inline-block;
+    width: 16px;
+    height: 16px;
+    margin-right: 0.4em;
+    vertical-align: -0.1em;
+
+    background-size: 16px 16px;
+    background: transparent url(img/location-blue.png) 0 0 no-repeat;
```
mhl: The link was appearing without the crosshair-icon, which @zarino worked out was because of these two lines being the wrong way round (otherwise the `background` overrides the background-size to be 0x0 again).

candidates/static/candidates/_finder.scss
```diff
@@ -181,6 +181,28 @@
   }
 }
 
+.geolocation-link {
+  display: block;
+  margin-top: -1.5em; // negate the margin-bottom on the main form input
```
mhl: This wasn't needed for me, and breaks the layout (the link overlaps the input box above)...

candidates/static/js/home_geolocation_form.js
```diff
@@ -0,0 +1,33 @@
+var constructGeolocationLink = function constructGeolocationLink($wrapper){
+    var geolocationIsSupported = false;
+    if ("geolocation" in navigator) {
+        geolocationIsSupported = true;
+    }
+
+    if(geolocationIsSupported) {
+        var t1 = $wrapper.attr('data-link-text') || 'Use my current location';
+        var t2 = $wrapper.attr('data-loading-text') || 'Getting location\u2026';
```
mhl: These strings need to be translated too; there's an example of how to use translated strings in Javascript here: https://github.com/mysociety/yournextrepresentative/commit/95f09cbc578e58

mhl:
Probably my biggest comments as this is at the moment are:
* I think all front pages should have the same link added
* There should be a couple of tests for GeoLocatorView

(I'll take another look after those are dealt with, if that's OK.)

I don't quite understand how the -1.5 margin didn't break the layout when @struan and @zarino's were trying this out - cached CSS perhaps?

candidates/static/candidates/_finder.scss
zarino: FWIW: If/when this branch and #674 get merged into master, we can upgrade `.geolocation-link` to use the `@high-dpi-background-image()` mixin, which has the rules the right way round.

candidates/static/candidates/_finder.scss
zarino: I have no idea why it looked ok on @mhl’s screen, but not on mine. As he says, could be a cache thing – but I'm unsure why the cache would have such a tiny effect with nothing else looking broken.

candidates/views/frontpage.py
```diff
@@ -12,6 +16,53 @@
...
 from ..forms import AddressForm
 
+
+class GeoLocatorView(View):
+    def get(self, request, **kwargs):
+        latitude = kwargs['latitude']
+        longitude = kwargs['longitude']
+
+        for election in Election.objects.current().prefetch_related('area_types'):
+            area_types = election.area_types.values_list('name', flat=True)
```
mhl: I don't think this is right, I'm afraid - `area_types` will just be set to the area type of the last election in the iteration. I think you'll need to do something like:

```python
        generation_with_types = defaultdict(list)
        for t in AreaType.objects.filter(election__current=True) \
            .values_list('election__area_generation', 'name'):
            generation_with_types[t[0]].append(t[1])
```

... and then do a MapIt query per generation, as for the UK postcode lookup:

https://github.com/mysociety/yournextrepresentative/blob/951fb25b627a6c1f623781eabd80919a9f2c552e/candidates/models/address.py#L34-L66

struan:
This will fix #605

PR #674, opened 2016-01-21
New header layout and candidate search box

zarino:
Part of #644.

The addition of a search box necessitated a reworking of the header layout. There is now a "masthead" bar with the site logo and user links in, and below it, a "nav" with site links and the search box.

Some of the repeated code for the search box background image and placeholder pseudo-element is spun out into two mixins, which live in the new mixins.scss file.

**Things that we should probably do before this goes live:**

- [ ] Styling for a “Candidate search results” page, that people will be taken to when they submit this new search form (the final part of #644).
- [ ] Sort of the styling of these new menus on narrow screens. (The YNR nav was never very mobile-friendly, but now it’s even less so.)

Here are some screenshots in the different country cobrands. (No screenshot of Yo Quiero Saber because it uses its own completely custom header, but I've checked, and none of my new styles conflict with YQS’s existing ones.)

screen shot 2016-01-20 at 15 57 56

screen shot 2016-01-20 at 15 59 32

screen shot 2016-01-20 at 16 00 53

Also, a screenshot of what it looks like when you're a logged in admin user with some photo approval notifications:

screen shot 2016-01-20 at 16 02 24


zarino:
This branch includes a handy `@include high-dpi-background-image()` mixin.

Once the branch is merged into master, we should check whether there are any other places we can use this mixin, so we're not constructing high dpi image rules manually in some places and via the mixin in others.

Eg: #665 is about to introduce a "Use my current location" cross-hairs icon to the homepage. That should use `@include high-dpi-background-image()` rather than constructing its own background-image styles.

struan:
Closed in favour of #678

PR #675, opened 2016-01-21
Fix email translations

struan:
Copy across the allauth templates so the translations are handled by our .po files


mhl:
For our email templates, I think I made each sentence a separate `blocktrans` translation and used the `blocktrans` tag's `trimmed` option, so that we didn't end up with newline-based email layout in the strings for translation - maybe worth doing the same with these since we're customizing them anyway?

Looks good, anyway.

PR #678, opened 2016-01-25
Redesign header, adding person search and index people for searching in ElasticSearch

### LINKED ISSUE #607: add a search box to look for a person's page
This would be useful in general (rather than having to use Google with a `site:` restriction) and the code would be a useful basis for #580 (or make it unnecessary for people who know the site really well).

### LINKED ISSUE #664: New header design to incorporate person search field
We need somewhere to add a search box so that people can look up a person.

This should also include a results page for the search output.

struan:
This PR obsoletes #674 but see there for details on the design changes.

It then adds Haystack hooks to index people in ElasticSearch and ties the Haystack search into the search box from #674 along with a search results page which has a first pass design.

Closes #607 and closes #664


candidates/templates/search/indexes/popolo/person_text.txt
```diff
@@ -0,0 +1,4 @@
+{{ object.name }}
+{{ object.given_name }}
+{{ object.family_name }}
+{{ object.additional_name }}
```
mhl: It'd be good if this included these fields from each alternative name in `other_names` as well

candidates/urls.py
```diff
@@ -204,6 +204,11 @@
         'pattern': r'^upcoming-elections',
         'view': views.UpcomingElectionsView.as_view(),
         'name': 'upcoming-elections'
+    },
+    {
+        'pattern': r'^search',
```
mhl: It's probably best to have a trailing `$` in the regexp to avoid catching any other `search`-prefixed paths in the future. (It might sound trivial, but we have had a similar bug in YNR in the past.)

candidates/search_indexes.py
```diff
@@ -0,0 +1,16 @@
+from haystack import indexes
+
+from popolo.models import Person
+
+
+class PersonIndex(indexes.SearchIndex, indexes.Indexable):
+    # FIXME: this doesn't seem to work for partial names despite what
+    # docs say
```
mhl: I don't know if this might be related, but the documentation suggests having the `EdgeNgramField` (for autocomplete in their example) as a new field: http://django-haystack.readthedocs.org/en/latest/autocomplete.html#step-1-setup-the-data

requirements.txt
```diff
@@ -45,6 +46,7 @@ Pillow==2.7.0
 psycopg2==2.5.4
 pyasn1==0.1.7
 pycparser==2.10
+pyelasticsearch==1.4
```
mhl: We're using Elasticsearch 0.9 on our servers, but `pyelasticsearch` [dropped support for that in v1.0](https://pyelasticsearch.readthedocs.org/en/latest/changelog/#v1-0-2015-01-23). SayIt restricts the `pyelasticsearch` version to >= 0.6, < 0.7. (Pombola uses Elasticsearch's `elasticsearch` Python package instead, pinned to `0.4.5` on the advice here: https://github.com/elastic/elasticsearch-py#compatibility )

candidates/search_indexes.py
struan: Yeah, that didn't seem to help and searching the internet turned up quite a few other people failing to get it to work so I punted on partial names after a couple of frustrating hours of trying various suggestions.

mhl:
I think the Travis failures are because the `.travis.yml` needs:

```
services:
  - elasticsearch
```

I left a comment about constructing the index name slightly differently - otherwise, assuming the tests then pass, I think this is good to go. :+1:  (I think as just one commit.)

mhl:
```<columbo>```Just one more thing...```</columbo>``` Could you add a test that adds a couple of people and checks one of them is returned from the search view too?

mhl:
:+1: Looks good (i just added a couple of tiny comments) - I assume you're going to squash out the commits that are really fixups before merging...

PR #680, opened 2016-02-01
Customize visible fields

### LINKED ISSUE #663: allow customization of the editable Popolo fields in the admin
# Background 

The fields that are crowd-sourced for candidates are essentially a flat list (although arranged into groups), e.g. for this part of the edit form:

fields

... there are various types of fields:

## 1. Simple fields

e.g. in the part of the form in that screenshot, "gender" and "date of birth" are "simple fields" - they directly map to a top-level field of [a Popolo person](http://www.popoloproject.com/specs/person.html). The fields in this category are [defined in `field_mappings.py`](https://github.com/mysociety/yournextrepresentative/blob/master/candidates/models/field_mappings.py#L1-L8)

## 2. Complex fields

e.g. in the part of the form in that screenshot, "homepage URL", and "wikipedia URL" these map to elements of arrays at the top level of [a Popolo person](http://www.popoloproject.com/specs/person.html). (e.g. the Wikipedia URL is an object in `links` with `note: wikipedia`; the value is stored under the key `url` in that object). The fields in this category are also [defined in `field_mappings.py`](https://github.com/mysociety/yournextrepresentative/blob/master/candidates/models/field_mappings.py#L10-L54)

## 3. Additional fields

e.g. in the part of the form in that screenshot, "important roles" and "profession" are additional fields. These are fields that aren't part of the Popolo standard, but which the site owners have requested. These are defined in the database rather than code, using the [`ExtraField`](https://github.com/mysociety/yournextrepresentative/blob/master/candidates/models/fields.py#L6-L26) and [`PersonExtraFieldValue`](https://github.com/mysociety/yournextrepresentative/blob/master/candidates/models/fields.py#L29-L36) classes.

# User needs

For each new jurisdiction we've customized YNR for there seems to be some need for customizing the fields that are crowd-sourced. The additional fields, since they're defined in the database, can be customized easily in the admin. However the simple fields and complex fields can't. We've been asked to do this several times, e.g.:

* Removing "honorific_suffix" for Argentina https://github.com/mysociety/yournextrepresentative/issues/470
* Removing honorific_prefix and honorific_suffix from Costa Rica (we did this with an [ugly conditional in the template](https://github.com/mysociety/yournextrepresentative/blob/91d1fa6f354e0435a7605579a3b09074105e717b/candidates/templates/candidates/person-edit.html#L62-L72))
* The request for YouTube links: https://github.com/mysociety/yournextrepresentative/issues/136
* There's interest in other fields that people might want to add: https://github.com/mysociety/yournextrepresentative/issues/240
* The popularity of various social networks is very different in different countries; we've been lucky so far to be able to get away with the selection we originally had.

So, the user story is something like:

```
As a site manager
I want to be able to customize the fields (like social network accounts) that are collected
So that the crowd's efforts are targetted at finding data that is useful
```
# Plan

This could be divided into two chunks of work; doing the simple fields and then the complex fields. Going by the work involved in introducing the customizable additional fields, (mostly in e6bf873364a5e2) this will probably involve:

- [ ] Creating a new model class to represent the fields that should be collected (presumably `SimplePopoloField` or `ComplexPopoloField` or similar) and generate migrations
- [ ] Update the serialization for the versions array (`get_person_as_version_data` and `revert_person_from_version_data`)
- [ ] Update the initializer of BasePersonForm to create form fields for all the the simple / complex fields in the database
- [ ] Update `update_person_from_form` in `candidates/models/popolo_extra.py`
- [ ] Update the `get_initial_form_data` method of `PersonExtra`
- [ ] Update the `candidates/person-edit.html` template and `UpdatePersonView` class
- [ ] Update the `candidates/person-view.html` template and `PersonView` class
- [ ] Add the new field models to the admin interface in `candidates/admin.py`
- [ ] Add `simple_popolo_fields` / `complex_popolo_fields` endpoints to the API
- [ ] Update the `candidates_import_from_live_site` command to mirror those classes
- [ ] Add a data migration to create the standard set of Popolo fields

struan:
This adds a table with a list of the simple PopoloFields that should be displayed on the site and then uses that list to determine which fields to display.

The set of fields is restricted to those that are part of the Popolo spec and are direct properties of the Person object.

NB: at the moment the Patronymic name isn't displayed anywhere because it's not clear how to do so.

This fixes #663

### DIFF
--- a/candidates/admin.py
+++ b/candidates/admin.py
@@ -5,7 +5,7 @@ from django.contrib import admin
 from django.core.urlresolvers import reverse
 from django.forms import ModelForm
 
-from .models import LoggedAction, PartySet, ExtraField, PersonExtraFieldValue
+from .models import LoggedAction, PartySet, ExtraField, PersonExtraFieldValue, SimplePopoloField
 # Register your models here.
 
 
@@ -58,7 +58,12 @@ class PersonExtraFieldValueAdmin(admin.ModelAdmin):
     form = PersonExtraFieldValueAdminForm
 
 
+class SimplePopoloFieldAdmin(admin.ModelAdmin):
+    list_display = ['name', 'label']
+
+
 admin.site.register(LoggedAction, LoggedActionAdmin)
 admin.site.register(PartySet, PartySetAdmin)
 admin.site.register(ExtraField, ExtraFieldAdmin)
 admin.site.register(PersonExtraFieldValue, PersonExtraFieldValueAdmin)
+admin.site.register(SimplePopoloField, SimplePopoloFieldAdmin)
--- a/candidates/forms.py
+++ b/candidates/forms.py
@@ -14,7 +14,7 @@ from django.contrib.sites.models import Site
 from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 
-from candidates.models import PartySet, parse_approximate_date, ExtraField
+from candidates.models import PartySet, parse_approximate_date, ExtraField, SimplePopoloField
 from popolo.models import Organization, Post
 
 class AddressForm(forms.Form):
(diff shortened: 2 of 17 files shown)


mhl:
Nice work, looks good :+1:

mhl:
The only issue I can see with this is that it makes the order of fields in the "new candidate" form even more eccentric than #682 is complaining about already:

fields-order

mhl:
@struan: after discussing this with @tmtmtmtm - let's merge this, and prioritize the form refactoring mentioned in #682 either at the end of this week, if you've got spare time, or at the start of next sprint

PR #689, opened 2016-02-05
allow multiple winners per post

### LINKED ISSUE #582: recording winners should work for posts which can have multiple winners
At the moment the code assumes that if you mark someone as the winner of an election for a post, everyone else lost, but some posts can have multiple people elected. We should support this (probably adding a property to Election `can_have_multiple_winners`).

mhl:
This adds an election-wide property allowing you to vary the number of
people who may be elected per post (previously this was implicitly 1).

Now you can set multiple candidates as having been elected to a post.

Ideally this should also have new tests to deal with the new cases possible
where people_elected_per_post is not 1, but we probably don't have time to
add them before the Costa Rican municipal elections this weekend.

Fixes #582


struan:
Looks good to me.

PR #695, opened 2016-02-12
Default to allowing CORS requests on API URLs

symroe:


mysite/settings/conf.py
```diff
@@ -414,6 +416,14 @@ def get_settings(conf_file_leafname, election_app=None, tests=False):
...
             },
         },
+
+        # CORS config
+        'CORS_ORIGIN_ALLOW_ALL': conf.get('CORS_ORIGIN_ALLOW_ALL', True),
+        'CORS_URLS_REGEX': conf.get('CORS_URLS_REGEX', r'^/api/.*$'),
+        'CORS_ALLOW_METHODS': conf.get('CORS_ALLOW_METHODS', (
+            'GET',
+            'OPTIONS',
+        )),
```
mhl: Since this is restricted to `/api/` and `GET` methods, I don't think I'd bother making this configurable from `general.yml`.  It would be easy for people to change them to insecure settings that we wouldn't recommend.

(Alternatively, if there's a good reason to make them configurable, it would be good to update `conf/general.yml-example` with an explanation of what each of the new options does.)

mhl:
Just one comment, otherwise this looks good, thanks! :+1:

mhl:
I've squashed these commits (and switched the django-date-extensions back to 1.1.0 rather than the equivalent git URL) and merged to master - that's https://github.com/mysociety/yournextrepresentative/commit/e6ce7508b1e7c72ff8e6d273a6cbad1ae358b98f

Many thanks!

PR #696, opened 2016-02-15
Email alerts for person updates

### LINKED ISSUE #606: [placeholder] add email alerts for updates to people (or any people of a particular party)
Required for Jamaica.

struan:
This adds the ability to sign people up to email alerts when:

 * A specific person is edited
 * A person in an area is edited
 * A person in a party is edited

The alerts can be sent either daily or hourly ( although if you are signed up to hourly alerts and daily alerts you will get the daily ones with the hourly ones )

It uses django-notifications to handle creating the notifications used for alerts, with a signal to create these attached to the save event of the LoggedAction model.

Alerts are created in the admin interface and require the django id of the object.

Fixes #606

### DIFF
--- a/alerts/admin.py
+++ b/alerts/admin.py
@@ -1,0 +1,10 @@
+from django.contrib import admin
+
+from .models import Alert
+
+
+@admin.register
+class AlertAdmin(admin.ModelAdmin):
+    list_display = \
+        ['user', 'target', 'action', 'frequency', 'last_sent', 'enabled']
+    list_filter = ('action', 'frequency', 'enabled')
--- a/alerts/management/commands/alerts_send_alerts.py
+++ b/alerts/management/commands/alerts_send_alerts.py
@@ -1,0 +1,114 @@
+from dateutil import parser
+from datetime import datetime, timedelta
+from django.core.management.base import BaseCommand
+
+from django.conf import settings
+from django.contrib.sites.models import Site
+from django.core.mail import send_mail
+from django.core.urlresolvers import reverse
+from django.template.loader import render_to_string
+
+from alerts.models import Alert
+
+
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        group = parser.add_mutually_exclusive_group()
+        group.add_argument(
+            '--hourly',
+            action='store_true',
+            help='Send hourly alerts'
+        )
+        group.add_argument(
+            '--daily',
+            action='store_true',
+            help='Send daily alerts'
+        )
+
+    def handle(self, **options):
+        if options['hourly']:
+            last_sent = datetime.utcnow() - timedelta(hours=1)
+            frequency = 'hourly'
+        elif options['daily']:
+            last_sent = datetime.utcnow() - timedelta(days=1)
+            frequency = 'daily'
+
+        alerts = Alert.objects.filter(
+            frequency=frequency,
+            last_sent__lt=last_sent
+        )
+
+        """
+        at the moment this means that if a user has multiple alerts set up
+        they will get all of them in the same email, even if they have set
+        up one alert to be daily and one to be hourly
+        """
+        for alert in alerts:
+            events = alert.user.notifications.unread() \
+                .order_by('action_object_content_type', 'action_object_object_id')
+
+            details = ""
+            current_object = ''
+            if events.count() > 0:
+                no_change_count = 0
+                change_count = 0
+                for event in events:
+                    if event.action_object != current_object:
+                        if current_object != '':
+                            details += "\n"
+                        details += 'Changes to {0}\n'.format(event.action_object)
+                        domain = Site.objects.get_current().domain
+                        path = reverse(
+                            'person-view',
+                            kwargs={'person_id': event.action_object.id}
+                        )
+                        protocol = 'http'
+                        if hasattr(settings, 'ACCOUNT_DEFAULT_HTTP_PROTOCOL') and \
+                            settings.ACCOUNT_DEFAULT_HTTP_PROTOCOL == 'https':
+                            protocol = 'https'
+                        details += "{0}://{1}{2}\n".format(
+                            protocol,
+                            domain,
+                            path
+                        )
+                        current_object = event.action_object
+                    if event.data:
+                        if event.data['changes'] is not None:
+                            timestamp = parser.parse(event.data['changes']['timestamp'])
+                            event.data['changes']['timestamp'] = timestamp
+                            diff = render_to_string('alerts/pretty_diff.txt', context=event.data)
+                            # remove extra blank lines
+                            diff = "\n".join([ll.rstrip() for ll in diff.splitlines() if ll.strip()])
+                            details += "\n{0}\n".format(diff)
+                            change_count += 1
+                        else:
+                            no_change_count += 1
+                    else:
+                        no_change_count += 1
+
+                if no_change_count == 1:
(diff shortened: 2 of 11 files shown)


alerts/admin.py
```diff
@@ -0,0 +1,11 @@
...
+
+from .models import Alert
+
+
+class AlertAdmin(admin.ModelAdmin):
+    list_display = \
+        ['user', 'target', 'action', 'frequency', 'last_sent', 'enabled']
+    list_filter = ('action', 'frequency', 'enabled')
+
+admin.site.register(Alert, AlertAdmin)
```
mhl: A small thing, but you can use `@admin.register` as a decorator on the `AlertAdmin` class now, which I think is a bit neater.

alerts/models.py
```diff
@@ -0,0 +1,139 @@
...
+        target_object_id=person.id
+    )
+
+    users_alerted = []
+
+    # when things are created they don't have an extra propery
+    # so we need to check to see if it exists
+    try:
+        changes = person.extra.version_diffs[0]
+    except:
```
mhl: It's almost always bad to have a bare `except` - it'll catch exceptions from typos in variable names or attributes, for example - there's no reason not to catch `ObjectDoesNotExist` here, I think.

alerts/models.py
```diff
@@ -0,0 +1,139 @@
...
+
+    alerts = Alert.objects.filter(
+        target_content_type=content_type,
+        target_object_id=person.id
+    )
+
+    users_alerted = []
+
+    # when things are created they don't have an extra propery
+    # so we need to check to see if it exists
```
mhl: I don't think this is correct, looking at: https://github.com/mysociety/yournextrepresentative/blob/58f71ab342a75c503ec3dc3c4ab955967363a5cd/candidates/views/people.py#L445-L472

alerts/models.py
```diff
@@ -0,0 +1,139 @@
...
+    verb = 'updated'
+    if created:
+        verb = 'created'
+
+    alerts = Alert.objects.filter(
+        target_content_type=content_type,
+        target_object_id=person.id
+    )
+
+    users_alerted = []
```
mhl: It's a bit nicer to make `users_alerted` a set, I think

alerts/models.py
```diff
@@ -0,0 +1,139 @@
...
+        if alert.user not in users_alerted:
+            users_alerted.append(alert.user)
+            notify.send(
+                person,
+                verb=verb,
+                recipient=alert.user,
+                changes=changes
+            )
+
+    # this doesn't handle people being removed from an area
```
mhl: Could you mark this as `FIXME` and create an issue for it? (Also people's party membership being changed away from the party for notification similarly won't be caught.)

alerts/management/commands/alerts_send_alerts.py
```diff
@@ -0,0 +1,116 @@
...
+                    if event.actor != current_actor:
+                        if current_actor != '':
+                            details += "\n"
+                        details += _('Changes to {0}\n').format(event.actor)
+                        domain = Site.objects.get_current().domain
+                        path = reverse(
+                            'person-view',
+                            kwargs={'person_id': event.actor.id}
+                        )
+                        details += "http://{0}{1}\n".format(domain, path)
```
mhl: We'll want the link to be `https` for some sites. I think the closest we can get in an admin command to telling if it should be `https` is the `NGINX_SSL` setting at the moment. Maybe we should remove the `FORCE_HTTPS_IMAGES` setting (no longer used) and replace it with an `SSL_BY_DEFAULT` setting to more generally specify that "in the absence of other information, assume this is using TLS/SSL" (in handling a request this isn't needed since we already know from `request.is_secure`, but it might be useful in management commands like this)

alerts/management/commands/alerts_send_alerts.py
```diff
@@ -0,0 +1,116 @@
...
+                            "There has been %(no_change_count)d change we don't have details of",
+                            "There have been %(no_change_count)d changes we don't have details of",
+                            no_change_count
+                        ) % {'no_change_count': no_change_count}
+
+                    details += "\n{0}\n".format(desc)
+
+                recipients = [alert.user.email]
+                has_sent = send_mail(
+                    _("Recent activity on YNR"),
```
mhl: You can use the sites framework to get the site name, which would be better than "YNR" here

alerts/management/commands/alerts_send_alerts.py
```diff
@@ -0,0 +1,116 @@
...
+                    details,
+                    settings.DEFAULT_FROM_EMAIL,
+                    recipients,
+                    fail_silently=False,
+                )
+
+                if has_sent:
+                    alert.last_sent = datetime.now(tz=pytz.utc)
```
mhl: Or just `datetime.utcnow()`?

alerts/models.py
```diff
@@ -0,0 +1,139 @@
...
+            content_type = ContentType.objects.get_for_model(party)
+            alerts = Alert.objects.filter(
+                target_content_type=content_type,
+                target_object_id=party.id
+            )
+
+            for alert in alerts:
+                if alert.user not in users_alerted:
+                    users_alerted.append(alert.user)
+                    notify.send(
```
mhl: It seems a bit strange to me that the "actor" is the candidate being edited or created. Given the django-notifications example:

> For example: justquick (actor) closed (verb) issue 2 (object) on activity-stream (target) 12 hours ago

... i'd have thought it's more natural for the actor to be the user making the change and the object to be the candidate. (That'd make it easier to understand the somewhat involved management command as well, I think.)

alerts/models.py
```diff
@@ -0,0 +1,139 @@
...
+                        person,
+                        verb=verb,
+                        recipient=alert.user,
+                        changes=changes
+                    )
+
+    # TODO: not sure this is the correct way to do this
+    # also need to make sure it's for a current election
```
mhl: Would `extra__election__current=True` be enough?

mhl:
Looks good - I think my comment about the actor / object change is probably the most significant thing to address.

alerts/admin.py
struan: On the other hand this is consistent with all the other admin files. I wonder if we should just make a ticket to swap them all. Or possibly just do it as part of this in a separate commit.

alerts/models.py
struan: Oh yes, this comment is out of date :( It was true when I was firing the signal on Person.save() but not now. I'll just remove this entire check as it should no longer be required.

alerts/admin.py
mhl: Sure, we can do that after this is dealt with - I'm not sure it's worth a ticket really.

alerts/management/commands/alerts_send_alerts.py
```diff
@@ -27,6 +28,8 @@ def add_arguments(self, parser):
         )
 
     def handle(self, **options):
+        lang = settings.LANGUAGE_CODE
+        activate(lang)
```
mhl: Since this is in a management command it doesn't actually matter, but in case someone copies this code to somewhere where it does matter, it might be nicer to use the `translation.override` context manager, e.g.

```python
from django.utils.translation import override
with override(settings.LANGUAGE_CODE):
    ...
```

mhl:
:+1: do go ahead and merge this

PR #705, opened 2016-02-16
Initial PCC Elections importer

symroe:


elections/uk_general_election_2015/management/commands/uk_general_election_2015_create_pcc_posts.py
```diff
@@ -0,0 +1,144 @@
...
+        election_defaults['organization'] = self.organization
+        election_defaults['name'] = election_defaults['name'].format(
+            area_name=mapit_area_data['name']
+        )
+        election_slug = self.format_election_slug(
+            self.format_org_slug(mapit_area_data['name']),
+            self.base_election_info,
+        )
+        print('Creating:', election_defaults['name'], '...',)
+        election, created = Election.objects.update_or_create(
```
mhl: This creates a new `Election` object for each PCC election. I can see that makes sense in one way, but I think it'll look a bit odd (e.g. on the "Elections" page) when we have a single Election for all of the Scottish Parliament regions but one for each police force region. I think I'd just make a single Election in `handle` which all of these Posts will be associated with.

elections/uk_general_election_2015/management/commands/uk_general_election_2015_create_pcc_posts.py
```diff
@@ -0,0 +1,144 @@
...
+            area.other_identifiers.update_or_create(
+                scheme=scheme,
+                defaults={'identifier': identifier},
+            )
+
+        AreaExtra.objects.get_or_create(base=area, type=area_type)
+        post, _ = Post.objects.update_or_create(
+            organization=self.organization,
+            area=area,
+            role='Police and Crime Commissioner for {area_name}'.format(
```
mhl: For consistency with what we did for Westminster, I think the role should just be "Police and Crime Commissioner". (This makes essentially no difference in practice at the moment, but it's easy to change...)

elections/uk_general_election_2015/migrations/0001_initial.py
```diff
@@ -0,0 +1,49 @@
...
+        old_id = area.identifier
+        try:
+            area = mapit_data[area.identifier]
+            new_id = mapit.format_code_from_area(area)
+
+            if old_id != new_id:
+                area.identifier = new_id
+                area.save()
+                print("Changed ID {0} to {1}".format(old_id, new_id))
+            for post in area.posts.all():
```
mhl: I was probably unclear when we discussed it in Slack, but I don't think there's any need to change the ID of the existing Westminster posts. In fact, it'll mean needing to add redirects for each of them since URLs with post IDs in them will have been bookmarked and indexed by Google. (URLs with area IDs in them are very unlikely to have been bookmarked, since they will only have been seen after the election, and then only in response to a POST request to the front page.)

So I think it's best to only change the Area IDs, not the post IDs. This will mean taking a bit of care in creating Post IDs in the future that don't collide with the existing Westminster post IDs, but I think that'll be pretty easy - just make sure new post IDs have a distinct single character prefix, say.

elections/uk_general_election_2015/management/commands/uk_general_election_2015_update_parties_from_ec_data.py
```diff
@@ -121,8 +121,8 @@ def parse_data(self, ec_parties_data):
                 party_extra = OrganizationExtra.objects.create(
                     base=party, slug=party_id
                 )
-                print("Couldn't find {0}, creating a new party {1}".format(
-                    party_id, party_name.encode('utf-8')
+                print(u"Couldn't find {0}, creating a new party {1}".format(
+                    party_id, party_name.encode('utf8').decode('utf-8')
```
mhl: The `.encode('utf8').decode('utf8')` can be completely left out, I think. And the `u"` prefix shouldn't be necessary since this file uses `from __future__ import unicode_literals`. What was the error that suggested this change? (It's worked in the past with non-ASCII names, I believe.)

elections/uk_general_election_2015/mapit.py
```diff
@@ -41,19 +60,20 @@ def get_areas_from_postcode(original_postcode):
...
-            ],
-            key=area_sort_key
-        )
+
+        result = []
+        for area in mapit_result['areas'].values():
+            if area['type'] in known_area_types:
+                code = format_code_from_area(area)
+                result.append((area['type'], code))
+        result = sorted(result, key=area_sort_key)
```
mhl: ha! I might be in a minority in finding the list comprehension version clearer, but the diff would have been smaller / easier to read if the new version was:

```python
        result = sorted(
            [
                (a['type'], format_code_from_area(a))
                for a in mapit_result['areas'].values()
                if a['type'] in known_area_types
            ],
            key=area_sort_key
        )
```

elections/uk_general_election_2015/management/commands/uk_general_election_2015_create_pcc_posts.py
tfgg: I thought we discussed that for PCC elections there would be one overall election, with each police area as a constituency, based on what the EC said about how it was regulated? Sorry if I wasn't clear.

elections/uk_general_election_2015/migrations/0001_initial.py
```diff
@@ -0,0 +1,49 @@
...
+logger = logging.getLogger(__name__)
+
+
+def mapit_ids_to_gss(apps, schema_editor):
+    Area = apps.get_model("popolo", "Area")
+    AreaType = apps.get_model("elections", "AreaType")
+    PostExtra = apps.get_model("candidates", "PostExtra")
+
+    initial_codes = list(set(AreaType.objects.values_list('name', flat=True)))
+    mapit_data = mapit.MapitLookup(initial_codes)
```
mhl: This will build a dictionary which is dependent on what `settings.MAPIT_BASE_URL` is at the time, so this migration would have to be run before the update to `MAPIT_BASE_URL` for the UK.

More broadly, I think I should probably argue a bit more (further to what I said in Slack) for including this mapping as an (admittedly large) dictionary literal in the migration instead of looking up MapIt. That gets around the following problems:

* At the moment the migration is dependent on a Django setting
* At the moment it's dependent on working connectivity for running (it's great to be able to work on trains or other places with poor connectivity)
* It's dependent on the implementation of the `MapitLookup` class not changing in any way in the future, which someone might well do without realising it'll break a migration
* It's dependent on the AreaType objects that happen to be in the database at the time

(We have had network dependent migrations in the past, like the one that migrated from PopIt, but I think it's arguable that that was a mistake. Non-deterministic migrations have caused me real problems in the past in several ways.)

That said, I really don't want to hold up your deployment because of this issue, because this will work fine at the moment - if you like, we can merge this as is, and I'll work on a different version later.

mhl:
Nice work, this almost all looks good - I think the two issues that I've commented on that are most worth addressing are the `Election` objects and `Post` ID issues.

One tiny thing (which probably isn't done right in the existing code anyway) is that forming the URLs for MapIt would be insensitive to whether the base MapIt URL has a trailing slash or not if they were concatenated with urljoin (`from django.utils.six.moves.urllib.parse import urljoin` for Python 3 compatibility). Again, that's just a "nice to have" level comment.

elections/uk_general_election_2015/management/commands/uk_general_election_2015_update_parties_from_ec_data.py
wfdd: There's another `encode` on line 118 that serves no purpose, if you wanna axe that too.  I've seen these encodes in a several other `commands`; they are likely an artefact of the author not having set their environment `LANG` to UTF-8.

elections/uk_general_election_2015/migrations/0001_initial.py
```diff
@@ -0,0 +1,43 @@
...
+                print("Changed ID {0} to {1}".format(old_id, new_id))
+
+        except KeyError:
+            print("No GSS code found for {}".format(area.identifier))
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('popolo', '0002_update_models_from_upstream'),
```
mhl: In case the migrations are run from scratch with an `ELECTION_APP` set such that data is pulled in from an existing PopIt instance, I think this should depend on: `('candidates', '0009_migrate_to_django_popolo')` in addition to 0002 from popolo.

elections/uk_general_election_2015/mapit.py
```diff
@@ -31,6 +40,19 @@ def area_sort_key(type_and_id_tuple):
...
+    code = None
+    if 'gss' in area['codes']:
+        code = str('gss:' + area['codes']['gss'])
+    elif 'ons' in area['codes']:
+        code = str('ons:' + area['codes']['ons'])
+    elif 'unit_id' in area['codes']:
+        code = str('unit_id:' + area['codes']['unit_id'])
+    elif 'police_id' in area['codes']:
+        code = str('police:' + area['codes']['police_id'])
+    return code
```
mhl: I think you might need to add `osni_oid` here, to deal with NIE areas:

* http://mapit.democracyclub.org.uk/area/23546.html
* http://mapit.mysociety.org/area/16972.html

candidates/views/areas.py
```diff
@@ -26,7 +26,7 @@ class AreasView(TemplateView):
     def get(self, request, *args, **kwargs):
         self.types_and_areas = []
         for type_and_area in kwargs['type_and_area_ids'].split(','):
-            m = re.search(r'^([A-Z0-9]+)-(\d+)$', type_and_area)
+            m = re.search(r'^([A-Z0-9]+)-([a-zA-Z0-9\:]+)$', type_and_area)
```
mhl: This needs to allow an `_` as well as the other characters in the identifier because of `unit_id`.

mhl: Actually, if you add the osni_oid codes too, they have hyphens in the identifiers, so maybe this should be:
```python
            m = re.search(r'^([A-Z0-9]+?)-([-a-zA-Z0-9\:_]+)$', type_and_area)
```
i.e. making the area type match non-greedy and adding `-` and `_` into the character class for the ID.

mhl:
I've left a comment about removing the output when migrating and everything's going fine.

Could you rename the `0001_initial` migration to something more descriptive? It's not really an initial migration - perhaps 0001_migrate_area_ids or something.

The other thing that occurs to me is that now that the UK installations of YNR are all past the "migrate from PopIt stage", we'd like to rename the `uk_general_elections_2015` app to just `uk`. This will stop new developers from automatically getting the lengthy and subject-to-network-failure "migrate from PopIt" migration and it better reflects the app's purpose now it's supporting 2016 too. It'll be much easier to do this before adding a migration to the app.  Do you mind if I introduce that change now, and you can rebase on top of that?

mhl:
Thanks for your patience in making all the revisions to this; I'm glad it's merged now :+1:

PR #711, opened 2016-02-23
Rename uk_general_election_2015 to uk

mhl:
The name of this ELECTION_APP is no longer accurate (since we're using
it for all the 2016 elections too).  It also make initial migrations
very long for people trying to set up the UK site for development, since
the 0009_migrate_to_django_popolo migration looks for that value and
tries to migration from PopIt.  That was useful during migration, but
now it's better for people to use the candidates_import_from_live_site
command.


struan:
Looks good to me.

PR #714, opened 2016-02-24
Create candidate form field ordering and design

### LINKED ISSUE #427: some polish would be nice for the create new candidate form
@zarino did a nice update to the candidate edit page so that the fields were helpfully grouped, but the new candidate form (both embedded and standalone) is just a `form.as_p`. It would be nice if this had a similar layout.

### LINKED ISSUE #682: the "source of information box" should come at the bottom of the "new candidate" form
At the moment, the "add a new candidate" form (e.g. from clicking "add a new candidate" on a post page) has the extra fields *after* the "what is the source of this information?" box

wrong-position-of-source-of-information-box

Generally, there was lots of nice work done by @zarino on improving the candidate edit page; it would be nice to port that to the "add new candidate" form.

struan:
This brings the design of the update form across to the create candidate form with some very minor tweaks to cope with the new context.

Fixes #682
Fixes #427


candidates/views/constituencies.py
```diff
@@ -171,6 +172,33 @@ def get_context_data(self, **kwargs):
             },
             hidden_post_widget=True,
         )
+        context['extra_fields'] = []
```
mhl: Since very similar code to this appears in three different views now, I think it would be good to factor it out (into a mixin, perhaps)

candidates/templates/candidates/_person_create_form.html
```diff
@@ -0,0 +1,141 @@
+{% load i18n %}
+
+  <form id="new-candidate-form" name="new-candidate-form" action="{% url 'person-create' election=election %}" method="post">
+    {% csrf_token %}
+    <h2>{% trans "Personal details:" %}</h2>
```
mhl: I have a similar comment here - I think a partial template could be factored out that's usable by both this template and person-edit.html

mhl:
If I click "add candidate" on an areas page, the form is missing the party select and hidden post input - this works fine when clicking on "add candidate" on a posts page, though.

symroe:
This looks good to me!

Issue #598, opened 2015-12-19
Group together similar elections


tfgg:
E.g. on https://edit.yournextmp.com/posts although "2016 London Assembly Election (Constituencies)" and "2016 London Assembly Election (Additional)" are separate elections for YNR's purposes as they have different election methods, it'd be helpful to group them together as "2016 London Assembly Election".

The group names could be higher-level election identifiers (#597), e.g. "gb-gla-2016-05-05" groups together "gb-gla-2016-05-05-a" and "gb-gla-2016-05-05-c".

mhl:
Let's just do a UK specific view and template override of this page; if it comes up in another country we can think about a more generic solution.

mhl:
We've come across another case where it would be nice if the data model supported grouping elections. @tfgg tells us that while the EC in the UK regulates the PCC election as a single election, but mayoral elections are regulated as separate elections. If we want the data model to follow how the state models elections (which @symroe has said is desirable) it would be helpful to be able to model that the mayoral elections are all the same kind of election, in some sense.

We could already do this by grouping the posts on that page by the post's `role` attribute and then the elections as a group underneath that. That might work rather well, actually.

jf1:
+1 from me on this, especially to improve pages that list all elections (currently rather chaotically) such as https://candidates.democracyclub.org.uk/posts

Can I add a +10 for sorting all these lists alphabetically too?

The current /posts page only seems useful for SEO - it's impractical for a human to use apart from searching within the page (for which a search facility may be a better option). If the various elections/posts were presented as an (initially collapsed) tree view then people would be able to navigate to their area of interest fairly quickly.

My grouping suggestion would be:
```
Devolved Assemblies and Parliaments
 - 2016 London Assembly Election
 - - Assembly Member for Barnet and Camden
 - - etc
 - 2016 Northern Ireland Assembly Election
 - 2016 Scottish Parliament Election (Constituencies)
 - 2016 Scottish Parliament Election (Regions)
 - 2016 Welsh Assembly Election (Constituencies)
 - 2016 Welsh Assembly Election (Regions)
Mayoral Elections
 - Bristol
 - etc
PCC Elections
 - all 
Metropolitan LAs
 - all metropolitan areas (definitely sorted alphabetically, possibly grouped by county)
Unitary LAs
 - all unitary areas (definitely sorted alphabetically, possibly grouped by county)
District LAs
 - all district areas (definitely sorted alphabetically, possibly grouped by county)
```

[...]

PR #715, opened 2016-02-24
Mayoral Election importer

symroe:
This is classed at WIP because I expect we'll need to figure out election groupings as part of this PR, or soon after.

Mayors are absolutely different elections, with their own IDs, however it's not very logical to list them differently on the site.  It could be that this is a very simple change to `posts.html`, or there might be something more complex.  mysociety/yournextrepresentative#598 talks about this more.

For now, I think this script imports the Mayors correctly, with different election IDs, making it technically "done"

### DIFF
--- a/elections/uk/management/commands/uk_create_mayoral_posts.py
+++ b/elections/uk/management/commands/uk_create_mayoral_posts.py
@@ -1,0 +1,155 @@
+from __future__ import print_function, unicode_literals
+
+from datetime import date
+
+from django.core.management.base import BaseCommand
+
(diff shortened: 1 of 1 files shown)


symroe:
Thanks Mark, I think the fixup commit addresses all your comments.  I'll squash it in and give the first commit a better name if you're happy with this.

mhl:
Looks good :+1:  - do go ahead

I made this PR with the fixes we discussed for the missing areas issues: https://github.com/mysociety/yournextrepresentative/pull/719 - if you could look over that too, that would be great.

PR #716, opened 2016-02-25
Jamaica site changes

### LINKED ISSUE #572: create a Jamaica theme
[placeholder issue]

### LINKED ISSUE #604: [placeholder] add a command to create posts and areas for Jamaica
We need to double-check with our partners whether they have managed to find shapefiles for constituencies. (If not, we can do a more limited geo-lookup at the province level anyway, and just show the representatives from the 4 or so constituencies within that province.)

However, we need this command in order to get a staging site / production site up for people to start using.

struan:
Theme for Jamaica plus text changes for it.

Also setup scripts for Parties and Areas.

Fixes #572 
Fixes #604


mhl:
:+1:

PR #717, opened 2016-02-25
Customize complex fields

### LINKED ISSUE #683: Allow customization of the complex Popolo fields
This is the second half of #663 and does the same thing for the complex fields such as links and contact details. See #663 for details of what to do.

struan:
This allows people to add custom fields based on contact details and links where the underlying data is stored in a related table and not as a direct property of the person.

The setup of them does require knowing what you are doing but there are hints in the admin interface to make it a bit easier.

It also has a migration to set up the standard fields that all existing YNR install have.

Fixes #683

### DIFF
--- a/candidates/admin.py
+++ b/candidates/admin.py
@@ -7,7 +7,7 @@ from django.forms import ModelForm
 
 from .models import (
     LoggedAction, PartySet, ExtraField, PersonExtraFieldValue,
-    SimplePopoloField, PostExtraElection
+    SimplePopoloField, ComplexPopoloField, PostExtraElection
 )
 
 
@@ -75,3 +75,9 @@ class PostExtraElectionAdmin(admin.ModelAdmin):
     list_filter = ('election__name', 'election__current')
 
     ordering = ('election', 'postextra__base__area__name')
+
+
+@admin.register(ComplexPopoloField)
+class ComplexPopoloFieldAdmin(admin.ModelAdmin):
+    list_display = ['name', 'label', 'order']
+    ordering = ('order',)
--- a/candidates/models/fields.py
+++ b/candidates/models/fields.py
@@ -42,6 +42,71 @@ class SimplePopoloField(models.Model):
     order = models.IntegerField(blank=True)
 
 
+class ComplexPopoloField(models.Model):
+    """
+    This model stores the name of the underlying relation, some details about
+    how it should be displayed ( label and field type ) and the details of
+    how to store the information in the generic relation.
+
+    The info_type_* properties are used to describe the key used to pull the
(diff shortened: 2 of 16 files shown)


candidates/models/popolo_extra.py
```diff
@@ -178,16 +178,18 @@ class PersonExtra(HasImageMixin, models.Model):
     images = GenericRelation(Image)
 
     def __getattr__(self, name):
-        if name in form_complex_fields_locations:
-            loc = form_complex_fields_locations[name]
+        # TODO: this does not seem optimal
+        field = ComplexPopoloField.objects.filter(name=name)
+        if field.exists():
+            field = field.first()
```
mhl: Similar to my comment on the other PR, you can save a query by doing

```python
field = ComplexPopoloField.objects.filter(name=name).first()
if field:
```

candidates/models/fields.py
```diff
@@ -42,6 +42,67 @@ class SimplePopoloField(models.Model):
...
+    )
+    info_type_key = models.CharField(
+        max_length=100,
+        help_text="Name of the field in the array that stores the type (note for links, contact_type for contacts, scheme for identifiers)"
+    )
+    info_type = models.CharField(
+        max_length=100,
+        help_text="Value to put in the info_type_key e.g. twitter",
+    )
+    old_info_type = models.CharField(max_length=100, blank=True)
```
mhl: It might be good to add some help here, just indicating that `old_info_type` is to support and `info_type` that was renamed, and usually isn't needed.

mhl:
Looks good to me, nice work :+1:

I'd still like to give it a bit of testing after these small changes and rebasing onto master, if that's OK - I've just reviewed by reading so far.

candidates/admin.py
```diff
@@ -75,3 +75,8 @@ class PostExtraElectionAdmin(admin.ModelAdmin):
     list_filter = ('election__name', 'election__current')
 
     ordering = ('election', 'postextra__base__area__name')
+
+
+@admin.register(ComplexPopoloField)
+class ComplexPopoloFieldAdmin(admin.ModelAdmin):
+    list_display = ['name', 'label']
```
mhl: Something that just occurred to me - it would be good if the ordering in the admin was by the `order` field. I think  you can just add `ordering = ('order',)` here to do that.

mhl:
If you go to an area page, and click "Add candidate" the complex fields are missing:

missing-complex-fields

Otherwise tthis all seems good.

candidates/admin.py
struan: I'm trying to decide if I'd find that more confusing than helpful.

PR #719, opened 2016-02-26
Handle missing areas from MapIt better

mhl:


symroe:
Yep, this looks geat :+1:

PR #720, opened 2016-03-01
Support UK IDs in cached counts app

symroe:


symroe:
Actually the ideal would be to have a per-install constant regex that would be used in all URL patterns that take an election ID.

I'm afraid I don't have enough time for that at the moment, but this change shouldn't make anything worse, and has the advantage of making the site work for UK elections.

struan:
This looks good to me.

PR #723, opened 2016-03-03
Further changes to cope with dots in election IDs

mhl:


struan:
Looks good to me.

Although I do think that we should probably aim to make sure that we just don't have . etc in the IDs by using the html_id from 8265677312688f9c52a40e6d90a25dd26535d6eb but better to be defensive :)

mhl:
Thanks for the review. I think I disagree about the `html_id` change, but we can discuss that at leisure another time :)

Issue #724, opened 2016-03-04
Unknown election slugs in the versions serialized JSON cause Person pages to 404


mhl:
This code is problematic:

* https://github.com/mysociety/yournextrepresentative/blob/e3faec3a414a022963967cb71c0354d2870ca6ea/candidates/diffs.py#L25

Since the election slug comes from the serialized JSON representing past versions of a person. Sometimes we have to rename election IDs, and this causes the page to 404.

In discussion with @symroe, we think the best thing to do is:

1. Add code to cope with an election not being found, and say it's an "unknown" election as an immediate workaround.
2. Take Pombola's `slug_helpers` application, make it into a standalone Python package and use its SlugRedirect model to maintain old election IDs which have been changed.

dracos:
Dunno if this is helpful, just to note SayIt uses django-sluggable (with a slight mySociety fork) to do automatic slug history (ie. when you update obj.slug and save, it creates the redirect, that kind of thing).

PR #725, opened 2016-03-04
Catch case where older election IDs were causing a 404

symroe:
The JSON for the diff stores the ID as it was at the time it was created.

This was causing 404 errors when `get_by_slug` was called on it


symroe:
Relates to https://github.com/mysociety/yournextrepresentative/issues/724

struan:
Looks good

PR #727, opened 2016-03-04
Various UK updates

symroe:
Sorry that this isn't very neat, but these are some of the various updates I've made to the UK site that don't fall into any other PR.


struan:
Looks good.

struan:
Although I notice that the first two commits in this are already merged in so possibly they should be removed before I merge this branch.

symroe:
Whoops, I've removed those two now – getting mixed up with maintaining a load of branches! :)

wfdd:
The changes in `_finder.scss` broke the front page styling of all the other sites.

Issue #691, opened 2016-02-06
allow you to customize the number winners on a per-post basis


mhl:
At the moment you can say that every post in a particular election may have N people elected to it (or for the number of winners to be unrestricted). However for the Costa Rican Regidor elections there may be a variable number of people elected depending on the post.

It would be nice (as @tmtmtmtm suggested in #582) if the data model supported overriding the election-wide number of winners per post. Note that this shouldn't just be a property of the post, though, since it's quite conceivable that the number of people elected per post is different between for that post in different elections. So probably this needs to be [an attribute on the ManyToManyField relationship](https://docs.djangoproject.com/en/1.8/topics/db/models/#extra-fields-on-many-to-many-relationships) between PostExtra and Election.

wfdd:
Intuitively, this would be a property of the constituency (area).

- Election has Area.
- Area has _x_ of Post in Election.

mhl:
Fixed in the linked pull request

PR #728, opened 2016-03-04
Customize winner count per post

struan:
Allow the number of winners in an election to be set on a per post level by using a custom M2M model between PostExtra and Election.

For #691

### DIFF
--- a/candidates/models/popolo_extra.py
+++ b/candidates/models/popolo_extra.py
@@ -518,6 +518,11 @@ class PostExtra(HasImageMixin, models.Model):
 
     candidates_locked = models.BooleanField(default=False)
     elections = models.ManyToManyField(Election, related_name='posts')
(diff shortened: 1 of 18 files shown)


candidates/models/popolo_extra.py
```diff
@@ -533,6 +537,13 @@ def short_label(self):
         return shorten_post_label(self.base.label)
 
 
+class PostExtraElection(models.Model):
+    postextra = models.ForeignKey(PostExtra)
```
mhl: It's probably not worth changing (in this PR anyway) but elsewhere the code generally uses `post_extra` rather than `post_extra` as a variable and attribute name.

mhl:
I've added some minor comments, only of which is serious (which does make me think that we should work on test coverage for more management commands).

Issue #729, opened 2016-03-04
"Elected"/"Not elected" on profile pages


tfgg:
On

https://candidates.democracyclub.org.uk/person/7455/sadiq-khan
https://candidates.democracyclub.org.uk/person/7457/peter-whittle

it shows their result for the 2015 UK GE. Obviously this is now misleading. It's not totally obvious what result should be shown, though.

symroe:
The fix in #732 has ben deployed now, so the links above no longer show the bug

PR #732, opened 2016-03-05
Remove election_to_show from the context

symroe:
It's not really clear why this was hardcoded, but it's not helpful now we have more than a single election.  

Later on, we need to think about this properly, so that it knows what elections are correct to show.

For now, let's just remove it.


symroe:
This fixes https://github.com/mysociety/yournextrepresentative/issues/729

PR #740, opened 2016-03-08
WIP: Fix too much data causing crash on update page

symroe:
When all the UK elections are in the database, the `/person/*/update` page becomes unusable due to the number select elements there are – there are over 2,000 posts and the party set dropdown is added for each election, meaning we get about 300,000 lines of HTML produced.

I suggest we only show the forms for elections the person is standing in, and the add a select2 widget for adding new elections.

I don't really know the best way to do this, but this PR does the initial work.  I would like to talk to others about how to move forwards with this!


struan:
A few initial thoughts on this are that it'd be nice if this was uk specific so either duplicate the _person_form.html template into the uk app or split out the candidacy section into a partial and have a uk version of that.

Likewise, possibly set is_standing as a property on the cons_form_fields and then using that in the template to determine if you should display the election would make this less uk specific.

One possible suggestion on how to make the JS work is to have an endpoint that just delivers up html from a partial with the various select options in and then inserts that in to the page. I know we do that for FixMyStreet in similar cases.

symroe:
Thanks!

I've just pushed a 2nd bash at this – it seems to work alright, but it's a little shonky.

symroe:
> Likewise, possibly set is_standing as a property on the cons_form_fields and then using that in the template to determine if you should display the election would make this less uk specific.

One of the problems is that iterating over all the elections was causing _a lot_ of database queries (over 5,000).

The new changes limit these queries to the elections that candidate is known to be standing in, meaning it's unlikely that there will be more than a few elections worth of queries per candidate.

There are now ~87 queries, vs over 5,000 before, and the UI is much cleaner.

The main thing I think is poor is the JavaScript – it's working, but isn't very nice, and I expct isn't really done in the "correct" way.  I'd like to talk to @zarino about this really!

elections/uk/templates/candidates/_person_form.html
```diff
@@ -0,0 +1,229 @@
...
+          </div>
+          <script type="text/javascript">
+          function showElectionsForm() {
+            $.getJSON("/api/current-elections/").done(function( data ) {
+              var select_data = $.map(data, function(obj, key) {
+                  return { id: key, text: obj.name + ' (' + obj.election_date + ')' };
+                })
+                $('#add_more_elections').select2({
+                data: select_data,
+              }).on('change', function(e) {
```
zarino: Looks like these two lines (`data:` and `}).on`) need to be indented 2 spaces further.

zarino:
I've just [added a commit](https://github.com/DemocracyClub/yournextrepresentative/commit/252f26d91f7310e1796b40c0c3f98d5b07be2275) that tidies up the new form a little bit. It now looks like:

candidacy

An obvious limitation is that you can't add a new election if you don't have JavsScript support enabled. Are we ok with that?

struan:
The result of that form doesn't look at all correct. It should only show the one party set on there and I imagine the standing dropdown should be first, probably default to Yes. And then the party selector and, if applicable, the party list one.

I think replacing the select2 setup call with

```
setUpPartySelect2s();
setUpPostSelect2s();
setUpStandingCheckbox();
updateFields();
```

and adding `.not('.select2-offscreen').not('.select2-container')` to the select2 setup calls in `setUpPartySelect2s` and setUpPostSelect2s` should make it all work. I'd add a commit for this but not commit privs to the DC github.

PR #748, opened 2016-03-17
WIP Display winning candidates at the top of area/cons pages

### LINKED ISSUE #692: Show the winner as the top result for the area page
When you go into http://tusrepresentanteslocales.co.cr and pick an alcalde (where the data is there) the winner seems to appear in random order. It would be good if this could be the first result on the page

For example:

screen shot 2016-02-08 at 09 05 29

and

screen shot 2016-02-08 at 09 08 38

struan:
Generate a list of winner candidates from the list of current
candidacies and use that to display a list of winners at the top of the
page.

Also skip display of winners in list of known candidates becuase we
don't want to repeat information

Fixes #692

### DIFF
--- a/candidates/templates/candidates/_candidates_for_post.html
+++ b/candidates/templates/candidates/_candidates_for_post.html
@@ -87,7 +87,11 @@
   {% for c, party in winners %}
     {% if forloop.first %}
     <div class="candidates__known">
-      <h3>{% blocktrans with post_label=post_data.label %}Winning candidates for <a href="{{ post_url }}">{{ post_label }}</a>{% endblocktrans %}</h3>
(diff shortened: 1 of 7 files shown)


mhl:
@struan Could you add a screenshot to the PR, possibly?

struan:
Still needs work on words, design and pluralisation - 

screen shot 2016-03-17 at 13 53 17

mhl:
Sorry, I was forgetting about the WIP pull request thing, so that request was a bit previous!

struan:
I think the next thing this needs is a once over from a designer to tidy up the display of winners.

struan:
And actually, this should probably now make use of the changes from #869

struan:
actually, this was done by #869.

PR #753, opened 2016-03-24
Add a country that just redirects to Democracy Club's site

mhl:
We want to redirect all requests to edit.yournextmp.com to Democracy
Club's candidates site (https://candidates.democracyclub.org.uk) from
now on.  This adds a new country we can switch edit.yournextmp.com to
use which will just redirect.


struan:
Looks good to me.

Issue #751, opened 2016-03-23
“Use my current location” says location not covered


andylolz:
![screen shot 2016-03-23 at 14 53 20](https://cloud.githubusercontent.com/assets/464193/13989157/0875a386-f107-11e5-8b55-84d1a9817ff0.png)

Newspeak House example: [Computer says no](https://candidates.democracyclub.org.uk/geolocator/51.5250159,-0.0713968), but postcode “E2 7DG” works and finds elections / candidates.

mhl:
There are two separate problems here - one is a logic error in `GeoLocatorView`, which there's a PR for here: https://github.com/mysociety/yournextrepresentative/pull/754

The other is that some of the elections have `area_generation` set to 22, but the DC MapIt installation only has generation 1:

```
ynr=# select slug, area_generation from elections_election order by area_generation DESC LIMIT 12;
            slug            | area_generation 
----------------------------+-----------------
 sp.r.2016-05-05            | 22
 gla.c.2016-05-05           | 22
 gla.a.2016-05-05           | 22
 2010                       | 22
 nia.2016-05-05             | 22
 sp.c.2016-05-05            | 22
 naw.r.2016-05-05           | 22
 naw.c.2016-05-05           | 22
 2015                       | 22
 mayor.bristol.2016-05-05   | 1
 mayor.liverpool.2016-05-05 | 1
 pcc.2016-05-05             | 1
(12 rows)
```

So they also need to be fixed with something like:

```
In [1]: from elections.models import Election

In [1]: Election.objects.filter(area_generation=22).update(area_generation=1)
Out[1]: 9
```

symroe:
Thanks – this seems to be working now.  @andylolz are you able to confirm?  (going to Newspeak not a requirement, I expect)

andylolz:
Funnily enough, I actually did test this at Newspeak! Yep – worked for me.

PR #754, opened 2016-03-24
Fix the GeoLocatorView for multiple MapIt generations

mhl:
The logic here was wrong, so it would think there were no results
at all if the first query to MapIt returned some, but the second query
returned none.

Combined with errors in the MapIt generation data in the DC install
of YNR this meant that any geolocation query would say "Your location
does not seem to be covered by this site".

Partial fix for #751


andylolz:
Oh yeah – good spot.

This looks good to me.

PR #763, opened 2016-03-31
Fix various problems with the "import from live site" command

mhl:


symroe:
Untested, but looks good to me

PR #764, opened 2016-03-31
Fix a test for whether an ImageExtra exists in CSV generation

mhl:
This test was broken - if an ImageExtra is missing, then
django.db.models.fields.related.RelatedObjectDoesNotExist is raised.
Catch the exception superclass for objects that don't exist instead.


symroe:
Tested on the live server and this works :+1:

PR #771, opened 2016-03-31
Add Python 3 cache folder to .gitignore

wfdd:
It would seem I remain the only Python 3 user. ;)


mhl:
Yes, looks like it :) Thanks for the pull request, I've merged that now.

PR #774, opened 2016-04-01
Display the party list link even if the people list isn't truncated

mhl:
It turns out to be nice to have a link to the party list even if the
list of candidates isn't truncated.


symroe:
:+1:

PR #786, opened 2016-04-04
Restore the ``u`` prefix in 'SOURCE_HINTS'

wfdd:
Unicode escapes inside 'bare' strings are not understood by
``xgettext``, the utility that creates translatable strings
from Python source code.  The backslash would be doubled in the
.po file; '\u2014' ≠ '\\\u2014'; and no translations
would then be found for 'SOURCE_HINTS'.


mhl:
:+1: I've tested this under Python 2 and 3, and it works fine. Thanks for the investigation into this...

PR #788, opened 2016-04-04
Mark up candidate list lock button for translation

wfdd:


candidates/models/popolo_extra.py
```diff
@@ -285,7 +285,7 @@ def age(self):
         if min_age == max_age:
             # We know their exact age:
             return str(min_age)
-        return "{0} or {1}".format(min_age, max_age)
+        return _("{0} or {1}").format(min_age, max_age)
```
mhl: Since you're changing this anyway, I think it might be nicer to use format variables that make the context clearer to a translator, e.g. it could be:

```python
return _("{min_age} or {max_age}").format(min_age=min_age, max_age=max_age)
```

wfdd: Done that now.

mhl:
Merged after squashing the fixup as: 9bdca4a884df9b5ef and 8c6c154f86e53

mhl:
Thanks!

PR #790, opened 2016-04-05
Emit lock form variables in areas view

wfdd:
This fixes being able to lock and unlock a post from an
areas view.


wfdd:
Though it still redirects you to the post afterwards, so there's that.

mhl:
:+1: Thanks! I've cherry-picked this as d55a0bf4089f96f3

Issue #785, opened 2016-04-04
Duplicate add candidate button at the top of candidate lists


wfdd:
Bit annoying having to scroll to the bottom to add another candidate, like [here](http://178.62.171.242:8001/areas/O06-805569).  (If there's no objections, I can open a PR.)

JoeMitchell:
Seconded.

wfdd:
Closed by #796.

PR #796, opened 2016-04-06
Repeat 'Add a new candidate' button at top of candidate list

wfdd:
The existing button will drop below the fold when the list
has grown sufficiently long.


wfdd:
Closes #785.

struan:
Looks good to me.

PR #798, opened 2016-04-06
Update API docs for fetching a post

andylolz:
I find django rest framework a bit baffling… But routes like `/api/posts/[post_id]/` seem to work, whereas `/api/posts/?extra__slug=[post_id]` just return the index (i.e. the `post_id` is ignored).

I think other bits on this page could be updated too… Mostly replacing “constituency” with “post”.


mhl:
Thanks, @andylolz - this documentation dated from a previous incarnation of the DRF-based API which handled post IDs in a broken way.  I've cherry-picked this as 1751a457cbdfec.

PR #800, opened 2016-04-06
Strip leading and trailing whitespace in submitted candidate data

### LINKED ISSUE #333: check whitespace is being trimmed from submitted data
@edent pointed out that some gender field entries have trailing whitespace (7174, 7297, 7107, 6423, 7306) - we should be calling `strip()` on input data.

wfdd:
This introduces a new class, ``StrippedCharField``, which
inherits from ``CharField`` and backports its Django 1.9 ``strip``
argument (see django/django@11cac1b).  It replaces all uses
of ``CharField`` in candidate forms with ``StrippedCharField``.

Closes #333.

### DIFF
--- a/candidates/forms.py
+++ b/candidates/forms.py
@@ -8,7 +8,7 @@ from .models.address import check_address
 
 from elections.models import Election
 
-from django import forms
+from django import forms, VERSION as django_version
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


struan:
Looks good to me

PR #803, opened 2016-04-07
Lose the Foundation presentational classes in 'image-review'

wfdd:
Previously, all of Foundation would be duplicated inside
'image-review' - some 250 KB - causing rendering issues.

### DIFF
--- a/moderation_queue/static/moderation_queue/css/crop.scss
+++ b/moderation_queue/static/moderation_queue/css/crop.scss
@@ -1,4 +1,5 @@
-@import "../../foundation/scss/normalize";
+$include-html-classes: false;
+
 @import "../../foundation/scss/foundation";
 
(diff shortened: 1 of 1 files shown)


zarino:
:+1: Looks good to me!

wfdd:
@zarino 

It looks like we've reverted #777 when merging?

mhl:
I'm not sure how this went wrong when Zarino merged, but I suspect it's something to do with 5a09867
being a merge commit. @wfdd - are you OK with me force-pushing 0ed6aa1a480 to master to fix that?

wfdd:
@mhl

Sure, whatever works.

mhl:
Great, I think that should be OK now.

wfdd:
Thanks :)

PR #813, opened 2016-04-09
Show big party list position numbers on post and area pages too

mhl:
I  made this change for the UK, to respond to a request from @symroe 
and @JoeMitchell - it means that the big numerals that were present on
the party list pages will now be used on the post and area pages too.

I was unsure about whether this would be generally welcomed by all
redeployers of YNR, but I think it's quite nice having the numbers on all
pages that display party lists. Do you have a view on that, @wfdd ?


wfdd:
This is actually something we wished for. :)

PR #819, opened 2016-04-10
Replace angle quotes for navigation with arrows

### LINKED ISSUE #791: Use actual arrows for navigation, not angle quotes
[These](https://en.wikipedia.org/wiki/Guillemet) are used as quote marks in several languages, including French and Greek.

wfdd:
Closes #791.


mhl:
:+1: Cherry-picked as 9bcf7c45885d8b4fd

PR #822, opened 2016-04-11
Put the parties with the most candidates at the top of the party drop-down

mhl:
A first pass at this - no tests yet, and it doesn't do the nice formatting
of right-aligning the candidate numbers as Zarino's sketch suggested,
but it's a start:

ranked-parties

### DIFF
--- a/candidates/models/popolo_extra.py
+++ b/candidates/models/popolo_extra.py
@@ -617,9 +617,54 @@ class PartySet(models.Model):
     def __str__(self):
         return self.name
 
-    def party_choices(self):
+    def party_choices_basic(self):
         result = list(self.parties.order_by('name').values_list('id', 'name'))
         result.insert(0, ('party:none', ''))
+        return result
+
+    def party_choices(self):
+        # For various reasons, we've found it's best to order the
+        # parties by those that have the most candidates - this means
+        # that the commonest parties to select are at the top of the
+        # drop down.  The logic here tries to build such an ordered
+        # list of candidates if there are enough that such an ordering
+        # makes sense.  Otherwise the fallback is to rank
+        # alphabetically.
+        candidacies_current_qs = Membership.objects.filter(
+            extra__election__current=True,
+            role=models.F('extra__election__candidate_membership_role'),
+            on_behalf_of__party_sets=self,
+        )
+        candidacies_ever_qs =  Membership.objects.filter(
+            role=models.F('extra__election__candidate_membership_role'),
+            on_behalf_of__party_sets=self,
+        )
+        minimum_count = settings.CANDIDATES_REQUIRED_FOR_WEIGHTED_PARTY_LIST
+        qs = None
+        if candidacies_current_qs.count() > minimum_count:
+            qs = candidacies_current_qs
+        elif candidacies_ever_qs.count() > minimum_count:
+            qs = candidacies_ever_qs
+        else:
+            return self.party_choices_basic()
+        result = [('party:none', '')]
+        parties_with_candidates = []
+        for party_tuple in qs \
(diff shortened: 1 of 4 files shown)


JoeMitchell:
Oooh, this is exciting!

jf1:
This will be such a time-saver as-is, so hope it goes live asap.

Agree with Joe that Zarino's formatting would be nice, but the lack of it shouldn't hold up deployment.

I do wonder if the UK version would benefit from the 'big 5' parties being sorted alphabetically, to prevent their order changing so people can quickly select from a stable list.
Or might the UK version suit having its lists sorted by 'most candidates in this country' to enable Plaid and SNP to be shown at the top of lists where they're likely to be present?

tfgg:
I was just thinking the same thing as jf1, it'd be good if the order could
be somewhat stable over time. Perhaps alphabetically sort in blocks of a
few hundred candidates?

wfdd:
On another point, when looking for 'Labour', the first result shouldn't be Democratic Labour.  I'd expect the results to be sorted according to their proximity score.  Perhaps the algorithm should weight proximity with the number of candidates?  I don't know if chunking is a good idea; at least, not when the list isn't being filtered.  We could have the top _x_ parties listed at the top, and then a little divider like you might have in a country dropdown, and everything else following that.

jf1:
I'd always been mindful that different countries might have a different number of 'big' parties, and that multiple chunks each with their own sort isn't ideal, so would suggest at most two chunks / one divider, with a variable so each local install can choose how many parties will be 'prioritised' and placed in alphabetical order at the top of the list.

All that said, it's not clear if there's an overall benefit from the split-list approach described above, so it's certainly not something I'd encourage to be prioritised.

The main reason I've come back to this thread is to comment after using experiencing the feature: As currently implemented, it is is such a (sanity- and) time-saver, so thanks @mhl.

candidates/models/popolo_extra.py
```diff
@@ -617,11 +617,56 @@ class PartySet(models.Model):
...
+        qs = None
+        if candidacies_current_qs.count() > minimum_count:
+            qs = candidacies_current_qs
+        elif candidacies_ever_qs.count() > minimum_count:
+            qs = candidacies_ever_qs
+        else:
+            return self.party_choices_basic()
+        result = [('party:none', '')]
+        parties_with_candidates = []
+        for t in qs \
```
struan: I'd quite like a more descriptive variable than t as off the top of my head I can't work out what it stands for.

mhl:
@wfdd I think that changing the item that's highlighted to be partly based on string proximity would need changes to Select2, unfortunately. I'm sure it could be done, but I think the reaction from people who've used this variant has been positive enough that I'd like to merge this anyway as an incremental improvement.

@tfgg Plaid Cymru and the SNP do come much closer to the top of the list with this variant, of course. I think the best solution to having them higher in Wales and Scotland is to have separate `PartySet`s for England, Scotland, Wales and Northern Ireland, rather than just Great Britain and Northern Ireland. We used the latter split because those are the two registers kept by the Electoral Commission, but as @andylolz has pointed out the EC also have "Field candidates in Scotland?" etc. fields so we could build the more specific party sets.

I'm not bothered by the stability of the top parties myself; having done quite a bit of candidate entry over the last week, my feeling is that experienced users will get the right party by typing a couple of characters after tabbing into the party select, and less experienced users will have to actually look at the drop-down anyway.

Anyway, as above I think we should merge this (@struan's reviewing it this morning)

candidates/models/popolo_extra.py
mhl: It was for "tuple" in my head, but fair point - I'll think of something better.

struan:
This looks good to me.

PR #828, opened 2016-04-13
Wrap long lines in multi-paragraph extra fields

wfdd:
``pre-line`` behaves the same way as ``normal``, save for
preserving new lines.  ``pre`` preserves new lines and repeated
spaces, and respects line length; it's what we'd use for
computer code.


wfdd:
I've bodged the bodge. ;)

mhl:
:+1: Thanks!

PR #829, opened 2016-04-14
Factor out repetitious test setup and remove sequence-based party & post creation

### LINKED ISSUE #594: factor out some common test setup into base classes
Lots of the tests' `setUp` methods create very similar objects (mostly mimicking the UK general election from 2015) - since there's so much repetition it would be good to factor out the most common setup into one or more test base classes.

mhl:
Fixes #594


struan:
Yay! This is much better than it was :)

PR #834, opened 2016-04-15
Fix "finder" styling on non-UK homepages

### LINKED ISSUE #831: Area finder styling broken
After #727, this is what the example app's (and all other apps') front page looks like:

Either the style changes should be factored out into the UK app or the finder mark-up should be brought in line with the UK's (which I think I'd quite prefer).

zarino:
3597b7e3022d9748422b45942077ca684b1a555f edited the `_finder.scss` styles in core, when really it should have been a UK-specific change.

This commit reverts those changes to core `_finder.scss`, and moves the `.finder__forms__container` and other related changes into a UK-specific override of the `_finder.scss` stylesheet.

Fixes #831.

# Before:

screen shot 2016-04-15 at 13 24 02

# After:

screen shot 2016-04-15 at 13 23 12

@mhl – could you give this a review? I've tested it locally by switching to the `uk` country, but all I see on the homepage is the location search form, rather than the extra stuff that’s visible on https://candidates.democracyclub.org.uk. I’m assuming there's some democracy club magic that I'm missing?

### DIFF
--- a/elections/uk/static/candidates/_finder.scss
+++ b/elections/uk/static/candidates/_finder.scss
@@ -110,9 +110,9 @@
     .header__hero {
       padding-top: 2em;
       padding-bottom: 4em;
+      text-align: center;
 
       .lead {
-        width: 50%;
         font-size: 1.5em;
         line-height: 1.4em;
         margin-top: 0.7em;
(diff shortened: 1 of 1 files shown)


mhl:
Thanks, @zarino - I've tried the front page in all the current countries, and then merged this into DC master to see if it works there too. Everything looked good :) :+1:

PR #839, opened 2016-04-15
Various template fixes

wfdd:
Attribute/dict access doesn't work inside ``blocktrans``.

### DIFF
--- a/candidates/templates/candidates/ordered-party-list.html
+++ b/candidates/templates/candidates/ordered-party-list.html
@@ -76,8 +76,8 @@
       {% endif %}
     {% empty %}
       <div class="no-candidates row">
-        <p class="medium-8 columns">{% blocktrans trimmed with post_label=post_data.label election_name=election_data.name %}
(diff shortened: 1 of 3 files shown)


wfdd:
For instance [[1]](https://candidates.democracyclub.org.uk/election/sp.r.2016-05-05/party-list/gss:S17000009/party:1):

> Oh no! We don’t know of any candidates from for Member of the Scottish Parliament for Central Scotland for the 2016 Scottish Parliament Election (Regions) yet.

That should be:

> Oh no! We don’t know of any candidates from **Beaconsfield Independent Electors Association** for Member of the Scottish Parliament for Central Scotland for the 2016 Scottish Parliament Election (Regions) yet.

wfdd:
I've pushed another commit in the same vein.  The site name was missing from the first sentence in `/tasks/` (see https://candidates.democracyclub.org.uk/tasks/).

wfdd:
Okay, I've pushed another unrelated fix so I'm renaming this to something suitably vague.

mhl:
:+1:  This all looks good, thank-you - I've rebased and then merged to keep the history simpler.

PR #807, opened 2016-04-08
Replace old PopIt gender in avatar selection with Popolo's

wfdd:
This fixes showing a silhouette of a woman in the absence of an
uploaded photo in candidate lists.

PR #842, opened 2016-04-15
Replace old PopIt gender in avatar selection with Popolo's

wfdd:


wfdd:
Same as with #807. I didn't think to look if it was the same elsewhere.

mhl:
:+1: Thanks, merged now

PR #844, opened 2016-04-16
Only include the record winner button on or after election day

### LINKED ISSUE #779: Only show “This candidate was elected!” buttons on or after the date of the election
Here’s the relevant bit, I suppose:
https://github.com/mysociety/yournextrepresentative/blob/0390d379/candidates/templates/candidates/_candidates_for_post.html#L135-L142

NB This template is overridden in the UK:
https://github.com/DemocracyClub/yournextrepresentative/blob/f3613d91/elections/uk/templates/candidates/_candidates_for_post.html#L133-L140

mhl:
These buttons add lots of visual clutter to the candidate lists, and are
only actually useful on election day or afterwards, so don't show them
until then.

Fixes #779


struan:
Looks good to me.

PR #848, opened 2016-04-17
Add ordering to extra fields

wfdd:


mhl:
Thanks @wfdd - this is great, but could you also add the new `order` parameter to the serializer of ExtraField for the API?

wfdd:
whoops, done that now.

mhl:
:+1: Looks good

mhl:
I've rebased this and then merged, hope that's OK

wfdd:
Yep, thanks.

PR #850, opened 2016-04-18
Sort person fields by their value of ``order`` by default

wfdd:
This obviates the need to use ``order_by``, which we'd
forgotten to do with the Popolo fields in the person view.


wfdd:
I didn't think this would require a migration but Django is of a different opinion.

struan:
Looks good to me

wfdd:
Thanks for merging these.

PR #865, opened 2016-04-25
Place footer links in nav-links list

wfdd:
The 'middot' character is the Greek semicolon and it's
more semantic this way anyway.

### DIFF
--- a/candidates/static/candidates/_footer.scss
+++ b/candidates/static/candidates/_footer.scss
@@ -19,6 +19,10 @@ body,
 
 .footer__links {
   margin-bottom: 0.5em;
+
+  .nav-links__item {
+    float: right;
+  }
 }
 
 .dc-logo {
--- a/candidates/static/candidates/_footer.scss
+++ b/candidates/static/candidates/_footer.scss
@@ -20,7 +20,7 @@ body,
 .footer__links {
   margin-bottom: 0.5em;
 
-  .nav-links__item {
+  .nav-links {
     float: right;
   }
 }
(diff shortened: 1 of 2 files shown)


wfdd:
before-after-footer

wfdd:
before-after-footer

candidates/static/candidates/_footer.scss
```diff
@@ -19,6 +19,10 @@ body,
 
 .footer__links {
   margin-bottom: 0.5em;
+
+  .nav-links__item {
```
zarino: Rather than floating each `.nav-links__item` individually, how about floating the entire `.nav-links` element to the right?

If you float the entire list rather than each item individually, you won’t have to reverse the order of the items in the HTML template, and as a bonus, the order of the links in the DOM will match the visual order which will be much more intuitive for people who tab through links with they keyboard or some other assistive device.

wfdd: Thanks, I've done that now.

wfdd: [IIRC, the reason I'd done that was the list items would wrap on the left, but that seems rather minor.]

zarino: Ah that is a good point. If we wanted to maintain a right-aligned wrap, then we could set the list items to be `{display: inline-block}` and then give the list itself `{text-align: right}`.

But as you say, it’s pretty minor. Maybe once we have a site that breaks onto a second line consistently, we should revisit.

zarino:
@mhl :+1: Looks great to me!

You'll want to auto squash before merging.

mhl:
Thanks, @wfdd and @zarino - that's in master now as 616c5ea380391f735f97374062fb1990f4378f5e

PR #867, opened 2016-04-26
On merging / reverting remove Memberships as well as MembershipExtras

### LINKED ISSUE #866: stale entries in popolo_membership left after merging
After merging old Membership objects still exist without their corresponding
MembershipExtra objects.

mhl:
This was a bug - the revert_person_from_version_data method,
which was used to implement merges of two candidates as well
as reverting them to an earlier version, deletes all existing
candidacies of a person and then recreates the ones that
should be there.  However, the code was only actually removing
the MembershipExtra objects and leaving the Membership objects
in the database.  These weren't shown on the user-facing site
because of missing the MembershipExtra objects, but would be
shown in the API output which was causing confusion.

This commit fixes that problem; a delete on the Membership
queryset deletes the corresponding MembershipExtras too.

Thanks to Sym Roe (@symroe) for spotting this and figuring
out that it related to merging.

Fixes #866

### DIFF
--- a/candidates/models/versions.py
+++ b/candidates/models/versions.py
@@ -133,9 +133,9 @@ def revert_person_from_version_data(person, person_extra, version_data):
         )
 
     # Remove all candidacies, and recreate:
-    MembershipExtra.objects.filter(
-        base__person=person_extra.base,
-        base__role=F('election__candidate_membership_role')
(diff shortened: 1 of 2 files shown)


struan:
This looks good to me.

PR #869, opened 2016-04-27
WIP: Display elected candidates on the candidate list page

zarino:
TODO:

- [x] @zarino: New "elected" icon files for each country
- [x] @zarino: `.candidates__elected` styling
- [x] @struan: Insert actual elected candidates into `candidates__elected` div in `_candidates_for_post.html`
    - (remember this also needs to handle party lists inside the `candidates__elected` div, just as `candidates__known` does)
- [x] @struan: Exclude elected candidates from `candidates__known`


struan:
So, the main issue I have is how to handle party lists inside the elected bit in the case where more members are elected than the maximum to display by default. At the moment it has the same behaviour as the normal page in that it shows N and the rest are on the party list page ( which will show elected and unelected member - possibly need to update this to hoist elected members to the top ).:

screen shot 2016-04-28 at 17 51 52

I'm not sure if for the elected case it shouldn't just display everyone elected. Also the 4 in the relates to the number elected rather than the total size of the list, but when you click through you get the whole list with elected and unelected people so I suspect wording needs to change there too.

It's also not clear that the 'See just this party list' is useful in the elected section. It probably is.

To summarise, the options seem to me to be:

a) leave it as is because mostly this will not be a problem and do something if it is
b) ignore the limit on how many candidates in a list to show in the elected section

Any thoughts on which of these, or if we want to do this another way?

wfdd:
I'm not sure if splitting the lists between elected and unelected candidates is such a good idea.  Because, essentially, the post view is a digital representation of the ballot paper, and the ballot paper - probably - should be reproduced as accurately as possible, even after the election has come to a close.  I'd opt for _repeating_ the winners compactly under the new 'Elected candidates' heading and restoring the complete list below under 'All candidates'.

mhl:
The more I think about it, the more I like @wfdd's suggestion here, i.e. that elected candidates should appear both in the usual presentation of candidates as before the election and in the special "winners" section at the top, but that the winners at the top of the page should be presented more compactly at the top.

I don't think we'll have time for much more work on this before results start coming in for the UK, however, so I think I'll probably make a pull request to the DC fork with essentially this change, but with some changes to address struan's comments:

> To summarise, the options seem to me to be:
> 
> a) leave it as is because mostly this will not be a problem and do something if it is
> b) ignore the limit on how many candidates in a list to show in the elected section

I think option (b) here makes more sense.

conf/general.yml-example
```diff
@@ -88,6 +88,13 @@ NEW_ACCOUNTS_ALLOWED: true
 # If this is set to false, then no edits of people are allowed.
 EDITS_ALLOWED: true
 
+# When candidates are marked as being elected, they're shown in a
+# special 'elected' section on post and area pages.  If this option is
+# false, they will shown both in that elected section and a complete
+# list of candidates below.  If it is turn, then it's as if the
+# elected candidates have been hoisted up to the elected section.
```
struan: s/turn/true

wfdd:
The election links in the 'known' and 'unelected' headings don't work on the DC site.

mhl:
Thanks, @wfdd - there's a pull request to fix that here: https://github.com/DemocracyClub/yournextrepresentative/pull/36

(That issue isn't in this topic branch.)

conf/general.yml-example
mhl: Thanks; fixed in 0caf4e2d84b47

struan:
What still need to be done to this PR for it to be done?

struan:
@wfdd We're going to merge this as is. I've created #929 for follow up work to provide a more compact winners list.

wfdd:
@mhl, @struan

When candidates aren't being hoisted, the full list's heading should be 'known' or 'all candidates'.  It now says 'unelected'.

PR #870, opened 2016-04-27
Fix searches featuring apostrophes

### LINKED ISSUE #862: Candidate search doesn't work for names with apostrophes
A search from https://candidates.democracyclub.org.uk/search finds nothing for names with apostrophes in, e.g. "O'Brien" (should find e.g. https://candidates.democracyclub.org.uk/person/11568/keith-charles-obrien) and "O'Mara" (should find e.g. https://candidates.democracyclub.org.uk/person/19405/margi-omara).

This also applies to the automated search for duplicates when reviewing newly entered candidates, e.g. re-entering "Margi O'Mara" failed to pick up any possible duplicates. (Given that, when this issue is fixed it would be good to check _all_ names with apostrophes in the DB for possible duplicates!)

struan:
The search uses the text field in the index which is generated by
Haystack and escapes the content as it goes in. Search terms were not
being escaped which meant that names with an apostrophe in them never
matched.

This escapes search terms so they should not match the values in the
text field in the search index.

Fixes #862

### DIFF
--- a/candidates/views/search.py
+++ b/candidates/views/search.py
@@ -1,5 +1,20 @@
+from django.utils.html import escape
+
 from haystack.generic_views import SearchView
+from haystack.forms import SearchForm
+
+
+class PersonSearchForm(SearchForm):
+    """
+    When Haystack indexed things into its sythesized text field
+    it HTML escapes everything. This means that we also need to
+    escape our search terms as otherwise names with apostrophes
+    and the like never match as the entry in the search index
+    is O&#39;Reilly and our search term is O'Reilly
+    """
+    def clean_q(self):
+        return escape(self.cleaned_data['q'])
 
 
 class PersonSearch(SearchView):
-    pass
+    form_class = PersonSearchForm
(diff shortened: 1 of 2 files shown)


candidates/views/search.py
```diff
@@ -1,5 +1,13 @@
+from django.utils.html import escape
+
 from haystack.generic_views import SearchView
+from haystack.forms import SearchForm
+
+
+class PersonSearchForm(SearchForm):
+    def clean_q(self):
+        return escape(self.cleaned_data['q'])
```
mhl: It'd be nice to have a comment here explaining why this is being HTML-escaped (i.e. because the indexed document based on `candidates/templates/search/indexes/popolo/person_text.txt` has HTML escapes

mhl:
Just a comment asking for a comment, :speech_balloon: otherwise this looks good :+1:

sjorford:
This doesn't seem to be working on candidates.dc - e.g. searching for "Fiona O'Donnell" doesn't find https://candidates.democracyclub.org.uk/person/839/fiona-odonnell

mhl:
@sjorford Thanks for pointing that out - this commit (390f7ded63d57efd4b8) is only in upstream, which is why the search fails on the DC fork. @symroe - I think if you cherry-pick that commit onto DC master, it should fix the problem there too, since:

    curl "http://localhost:9200/_search?q=Fiona%20O%26%2339;Donnell&pretty=1"

... does find the candidate.

Issue #873, opened 2016-04-28
fix bug with broken parsing of JSON patches


mhl:
If you just reorder items in an 3 item array, then the JSON patch that is generated might be something like the following:

```
=======
op: add
path: /other_names/0
value: {u'note': u'', u'name': u'Joey', u'end_date': u'', u'start_date': u''}
=======
op: remove
path: /other_names/3
```

The idea is that these operations are applied in order; first the new item is inserted at the start, then the fourth item, at index 3, is removed.

However, the JSON patch isn't used like that in `candidates/diffs.py` - it tries to look up `other_names/3` in the original data, which gives this error:

```
JsonPointerException: index '3' is out of bounds
```

Instead, it should apply the earlier operations before trying to parse the later ones.

This error is causing some person pages to error.

mhl:
There's a failing test for this in the [873-broken-jsonpatch-parsing](https://github.com/mysociety/yournextrepresentative/tree/873-broken-jsonpatch-parsing) branch (PR here: https://github.com/mysociety/yournextrepresentative/pull/874)

PR #874, opened 2016-04-28
Fix broken JSON patch handling

mhl:
Fix for: #873


struan:
I think this looks ok.

PR #875, opened 2016-04-28
Group elections more logically on pages that list all elections / posts

### LINKED ISSUE #598: Group together similar elections

### LINKED ISSUE #863: Current and past elections appear in random order in /numbers/
It would seem sensible for current elections to appear first.

mhl:
These commits change the following pages:

* `/numbers`
* `/posts`
* `/help/api`

... so that elections and posts are groups by:

* Group by `current=True`, then `current=False`
  * Group by `for_post_role` (ordered alphabetically)
    * Order election by election date (new to old) then election name
      * Order posts by post label

When there are a lot of elections, this should make their structure
more obvious.  Grouping by `for_post_role` means that the elections
representing constituency and regional elections in the Scottish
Parliament Elections come together.

This also adds a link to the all candidates CSV file, which was
missing.

Fixes #863 
Fixes #598 

new-posts-page

### DIFF
--- a/candidates/templates/candidates/posts.html
+++ b/candidates/templates/candidates/posts.html
@@ -13,18 +13,24 @@
 
   <p>{% trans "Follow one of the links below to see the known candidates for that post:" %}</p>
 
-  {% for election_data in all_posts %}
-    {% with election=election_data.election posts=election_data.posts %}
-        <h2>{{ election.name }}</h2>
-
-        <ul>
-          {% for p in posts %}
-            <li>
-             <a href="{% url 'constituency' election=election.slug post_id=p.slug ignored_slug=p.short_label|slugify %}">{{ p.base.label }}</a>
-            </li>
-          {% endfor %}
-        </ul>
-    {% endwith %}
+  {% for era in elections_and_posts %}
+    {% if era.current %}
+      {% for role_data in era.roles %}
+        <h2>{{ role_data.role }}</h2>
+        {% for election_data in role_data.elections %}
+          {% with election=election_data.election %}
+            <h3>{{ election.name }}</h3>
+            <ul>
+              {% for p in election_data.posts %}
+                <li>
+                 <a href="{% url 'constituency' election=election.slug post_id=p.slug ignored_slug=p.short_label|slugify %}">{{ p.base.label }}</a>
+                </li>
+              {% endfor %}
+            </ul>
+          {% endwith %}
+        {% endfor %}
+      {% endfor %}
+    {% endif %}
   {% endfor %}
 
 {% endblock %}
--- a/elections/models.py
+++ b/elections/models.py
@@ -87,3 +87,124 @@ class Election(models.Model):
 
     def __str__(self):
         return self.name
+
+    @classmethod
+    def group_and_order_elections(cls, include_posts=False):
+        """Group elections in a helpful order
+
+        We should order and group elections in the following way:
+
+          Group by current=True, then current=False
+            Group by for_post_role (ordered alphabetically)
+              Order election by election date (new to old) then election name
+
+        If the parameter include_posts is set to True, then the posts
+        will be included as well.
+
+        e.g. An example of the returned data structure:
+
+        [
+          {
+            'current': True,
+            'roles': [
+              {
+                'role': 'Member of Parliament',
+                'elections': [
+                  {
+                    'election': <Election: 2015 General Election>,
+                    'posts': [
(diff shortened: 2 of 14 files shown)


elections/models.py
```diff
@@ -87,3 +87,116 @@ class Election(models.Model):
...
+            qs = qs.prefetch_related(
+                models.Prefetch(
+                    'posts',
+                    PostExtra.objects.select_related('base') \
+                        .order_by('base__label'),
+                )
+            )
+        last_current = None
+        for election in qs:
+            current_index = 1 - int(election.current)
```
struan: A comment might be nice here just to explain this.

struan:
Feels like 0e9750b should be squashed into the previous commit.

candidates/templates/candidates/posts.html
```diff
@@ -13,18 +13,24 @@
...
+                <li>
+                 <a href="{% url 'constituency' election=election.slug post_id=p.slug ignored_slug=p.short_label|slugify %}">{{ p.base.label }}</a>
+                </li>
+              {% endfor %}
+            </ul>
+          {% endwith %}
+        {% endfor %}
+      {% endfor %}
+    {% endif %}
   {% endfor %}
```
struan: This sort of feels like it should be a partial given it's in two places already.

struan:
Tiny comments that aren't all that important but otherwise this looks good.

jf1:
Yay, thanks @mhl!

Looks like this will just group posts within the same election - is that correct and if so any plans to implement or support for further grouping at a higher level, eg so DC could group all this year's Mayoral elections, Local Authority elections, etc?

symroe:
Sorry I missed the review on this.

One thing I'd like to add is that by-elections have two groupings in the election ID, like `local.braintree.witham-south.2016-05-05`.  This means the migration should look like this:

```
diff --git a/elections/uk/migrations/0003_adjust_roles_for_grouping.py b/elections/uk/migrations/0003_adjust_roles_for_grouping.py
index 027cce1..969f8d1 100644
--- a/elections/uk/migrations/0003_adjust_roles_for_grouping.py
+++ b/elections/uk/migrations/0003_adjust_roles_for_grouping.py
@@ -9,7 +9,7 @@ from django.db import migrations
 def adjust_roles_for_grouping(apps, schema_editor):
     Election = apps.get_model('elections', 'Election')
     for election in Election.objects.all():
-        if re.search(r'^local\.[^.]+\.2016', election.slug):
+        if re.search(r'^local\.[^.]+(\.[^.]+)?\.2016', election.slug):
             election.for_post_role = 'Local Councillor'
             election.save()
         if re.search(r'^mayor\.[^.]+\.2016', election.slug):
```

Given that I expect this hasn't been run anywhere yet, should we update this migration or should I make a new one?

mhl:
@jf1 The level of grouping above election is the "role", which for Mayors will all be "Mayor", for local elections it'll be "Local Councillor", so this should do what you want as far as grouping goes. (The presentation is very basic still—just different levels of heading—but it's a good basis for a better version, e.g. hiding / showing who sections, or whatever.

jf1:
These pages are so much easier to look at now, and each group's elections are sorted alphabetically too :)

Some way to hide/expand each group's elections would help, and even better if collapsed groups in /numbers could display the total number of candidates, posts and unlocked posts within each group.

I previously suggested a sortorder value so eg Mayoral and PCC could go above Locals, but if they were all collapsed on initial display (ie so all top-level groups could be seen without scrolling down) then it wouldn't be necessary - or beneficial - so leaving them in alphabetical order seems best.

ps @symroe there are several current by-elections and I haven't noticed any problems with them tonight.

wfdd:
This got rid of the `{% empty %}` clause in `reports.html`, so there's no longer a message below an empty current or past election list.

mhl:
@symroe Feel free to modify the migration in this case

mhl:
Oops, the `{% empty %}` clause was restored later 52ff24317d7ffbeff6e772719b7a9198c46a2f6e - thanks, @wfdd

PR #882, opened 2016-04-30
Move to libsass and pipeline 1.6.8

### LINKED ISSUE #707: switch to using the libsass Python package instead of the sass gem
This was suggested by @wfdd on https://github.com/mysociety/yournextrepresentative/pull/539

> You could try using libsass, which would also rid you of your sole Ruby dependency.
> 
> In requirements.txt, add
> 
> ```
> libsass==0.10.0
> ```
> 
> and in conf.py, replace
> 
> ```
> 'PIPELINE_SASS_ARGUMENTS': '--trace',
> ```
> 
> with
> 
> ```
> 'PIPELINE_SASS_BINARY': '/usr/bin/env sassc',
> ```
> 
> It works for me, for whatever that's worth.

This seems like a good idea to me. I'm going to merge the Sass upgrade first, however.

mhl:
This provides a massive performance boost for dev environments

(I'm making a pull request from this commit of Sym's: we need to test
whether libsass and this django-pipeline upgrade work properly on
mySociety's servers, but I love this change and I'm keen to merge it -
development is *so* much nicer.)

Fixes #707

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,6 @@ addons:
   postgresql: "9.1"
 
 install:
-  - gem install --no-ri --no-rdoc compass -v 0.12.7
-  - gem install --no-ri --no-rdoc sass -v 3.2.19
   # Make sure pip is at the latest version:
   - pip install -U pip
   # Now install the rest of the required Python packages:
--- a/bin/pre-deploy
+++ b/bin/pre-deploy
@@ -56,19 +56,6 @@ fi
 # get the database up to speed
 ./manage.py migrate --noinput
 
-# Install gems in order to compile the CSS
-mkdir -p "../gems"
-export GEM_HOME="$(cd ../gems && pwd -P)"
-export PATH="$GEM_HOME/bin:$PATH"
-if (sass --version | grep '3\.4\.21') > /dev/null
-then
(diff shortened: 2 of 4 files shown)


mhl:
(@dracos reviewed this and we discussed it in IRC)

wfdd:
@mhl: I assume [[1]](https://github.com/mysociety/yournextrepresentative/blob/4197fb2f436588e49863f7ef4b3f308db604f8ec/conf/general.yml-example#L27) and [[2]](https://github.com/mysociety/yournextrepresentative/blob/c492dd2752c40ee9905e64c340bb6f7ed0ef3b16/mysite/wsgi.py#L43-L51) can go too?

mhl:
> @mhl: I assume [[1]](https://github.com/mysociety/yournextrepresentative/blob/4197fb2f436588e49863f7ef4b3f308db604f8ec/conf/general.yml-example#L27) and [[2]](https://github.com/mysociety/yournextrepresentative/blob/c492dd2752c40ee9905e64c340bb6f7ed0ef3b16/mysite/wsgi.py#L43-L51) can go too?

Thanks, I've removed them in 2038c565e6da994993ea050

PR #905, opened 2016-05-06
Log the rejection reason for photo moderation operations

### LINKED ISSUE #896: Log photo rejection reason
A copy of the rejection email's sent to the admins but also storing the rejection reason in the database so it can be reviewed from within the app would be an improvement.

struan:
This adds a note field to the LoggedActions model and uses that to store the rejection reason when uploaded photos are rejected during moderation.

It also adds a migration to update the help text of the complex fields because there was a mismatch there.

Fixes #896

### DIFF
--- a/candidates/admin.py
+++ b/candidates/admin.py
@@ -23,7 +23,7 @@ class LoggedActionAdmin(admin.ModelAdmin):
     list_filter = ('action_type',)
     list_display = ['user', 'ip_address', 'action_type',
                     'popit_person_new_version', 'person_link',
-                    'created', 'updated', 'source']
+                    'created', 'updated', 'source', 'note']
     ordering = ('-created',)
 
     def person_link(self, o):
--- a/candidates/migrations/0030_update_complexfields_help_text.py
+++ b/candidates/migrations/0030_update_complexfields_help_text.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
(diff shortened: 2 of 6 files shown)


wfdd:
Looks like another migration got caught up in this?

struan:
I deliberately added that migration in because it looks like it was missed earlier and it without it I was having to edit the makemigrations output.

wfdd:
Oh right, sorry.

mhl:
It would be good to have the missing migration in a separate commit, I think (with the commit message referencing the commit it should have been included in). Otherwise this looks good to me :+1:

mhl:
The test failure would be fixed by 3e6b734c79f3017a5b4b0c1bc102fbb8adc3dea4, I think

mhl:
One other thing about this - I think should be simple to update one of the existing tests to check that the reason is recorded in the LoggedAction in the case of rejecting a photo, so it would be good to do that too.

mhl:
Sorry, @struan - one last comment, I promise! I think the commit message for https://github.com/mysociety/yournextrepresentative/pull/905/commits/e3787996c05cea4acf0011b2a11d9e72c79827bd should indicate that it's fixing a mismatch between the model and migrations. (The commit that added the migration to create the table didn't match the model in `candidates/models/fields.py` in the same commit, I guess because the models were fixed-up without regenerating the migration.)

The way the commit message is phrased at the moment I'd expect the commit to have the change both in the models and migrations, and really it's only making sure the migrations are back in sync with the model.

PR #914, opened 2016-05-10
Improvements to CSV generation

### LINKED ISSUE #891: the all candidates CSV file should include a line per election
At the moment the "all candidates" CSV file generated by `candidates_create_csv` doesn't work well when candidates can stand in multiple elections. If a candidate is standing in multiple elections they should really have multiple rows - much of the data will be the same in each row, but there are candidacy-specific fields.

### LINKED ISSUE #911: generate CSV files just with results of current elections
This is key to make it easy to populate EveryPolitician with data from YNR.

mhl:
This PR introduces a number of improvements to CSV generation:

* The time taken to generate CSV files (tested on my laptop) has reduced from about 30 minutes to 3 minutes.
* CSV files that only include the successfully elected candidates are also generated. Fixes #911 
* If a person has been a candidate for more than one post, they appear once for each post Fixes #891
* Useful extra columns are added to the CSV output.
* There are now tests that check that the number of queries required for CSV generation are constant per election.
* Sorting is improved; first names are also considered in sorting, and the sorting for the "all candidates" file is different from the election-specific files (alphabetical by name and grouped by posts respectively)

And an incidental improvement:

* Only a single query on ComplexPopoloField is used when accessing attributes of a PersonExtra.


candidates/models/popolo_extra.py
```diff
@@ -443,100 +443,110 @@ def create_from_form(cls, form):
...
-            twitter_user_id = self.base.identifiers.get(
-                scheme='twitter',
-            ).identifier
-        except Identifier.DoesNotExist:
+                continue
+            if not m_extra.election:
+                continue
+            expected_role = m.extra.election.candidate_membership_role
+            if election is None:
+                if expected_role == m.role:
```
struan: I assume this is to avoid things like not standing candidates? A bit of a comment to explain this would certainly be helpful.

candidates/models/popolo_extra.py
```diff
@@ -443,100 +443,110 @@ def create_from_form(cls, form):
...
+            elected = candidacy_extra.elected
+            elected_for_csv = ''
+            image_copyright = ''
+            image_uploading_user = ''
+            image_uploading_user_notes = ''
+            proxy_image_url_template = ''
+            if elected is not None:
+                elected_for_csv = str(elected)
+            mapit_identifier = None
+            for identifier in post.area.other_identifiers.all():
```
struan: I sometimes wonder if we should find a way to extract this pattern out to somewhere as it's used in quite a few places. 

Possibly a get_key_from_collection helper method.

candidates/views/constituencies.py
```diff
@@ -202,15 +202,15 @@ def get(self, request, *args, **kwargs):
...
+            d for me in
             MembershipExtra.objects \
                 .filter(
                     election=self.election_data,
                     base__post=post
                 ) \
                 .select_related('base__person') \
                 .prefetch_related('base__person__extra')
+            for d in me.base.person.extra.as_list_of_dicts(self.election_data)
         ]
```
struan: I totally do not understand this list comprehension :(

mhl: @struan: Fair point, I'll unwrap this one. Nested list comprehensions are horribly difficult to write as well as read, I think, largely because the order of them seems completely the wrong way round to me, anyway - e.g. surely the first (wrong) example here is how you'd want to write it? http://stackoverflow.com/a/8050398/223092

mhl: Fixed in 1a8aa2070191708d37caf2ae767f3d8a847e27a1

candidates/models/popolo_extra.py
mhl: Whenever there's a condition on the `role` of a membership, it's basically guarding against people using the database to store memberships other than candidacies. The YNR codebase doesn't actually do that at the moment, but it was the intention behind `winner_membership_role` for example (i.e. that when someone's marked as a winner we also create their actual membership of the election organization's)

I can see someone adding other django-popolo-using Django applications to their site too, which might well create other types of memberships, so it's thinking of that as well.

Although I could add a comment here, I'm not sure it's the best place, since the same comment would apply basically anywhere we filter on `candidate_membership_role`. I'm not sure where the best place to document it is, though - perhaps we need a "how YNR uses django-popolo" file in `docs/` or something?

candidates/models/popolo_extra.py
mhl: Yeah, one thing that'd be nice about that is that we can document *why* it's being used in just that place. Every time we use iteration + test rather than a filter I feel like it needs explanation.

struan:
A few small comments aside this all seems to work and looks good.

I do note that if they remove some of the standard complex fields ( e.g. twitter_username ) then this crashes.

mhl:
Thanks, @struan. I've created [a ticket for the issue that CSV generation doesn't get its list of columns from `SimplePopoloField` / `ComplexPopoloField`](https://github.com/mysociety/yournextrepresentative/issues/915). As we discussed in IRC, as well, I think it's best to put off dealing with [this comment](https://github.com/mysociety/yournextrepresentative/pull/914#discussion_r62666975) until after these changes are deployed.

jf1:
@mhl Is is possible to also add a **votes** column to the [all candidates csv](https://candidates.democracyclub.org.uk/media/candidates-all.csv)?

mhl:
@jf1 Once the PR is closed, could you create a new ticket for that kind of request? Otherwise it's likely to be overlooked.

jf1:
Apologies, will do.

candidates/templates/candidates/results.html
```diff
@@ -0,0 +1,78 @@
...
+{% endblock %}
+
+{% block content %}
+
+<div class="help-api">
+
+<h2>{% trans "RSS/Atom feeds of results" %}</h2>
+
+{% blocktrans trimmed %}
+  There is an atom feed of recent election results available in two versions:
```
wfdd: @mhl

A little to the party but I think this should've been placed inside a `<p>` element.  Also, 'atom' should probably be capitalised, to 'Atom'.

PR #918, opened 2016-05-11
WIP: Improve date parsing in candidate forms

### LINKED ISSUE #907: Partial ISO date raises ValidationError
```
Internal Server Error: /election/vouleftikes-ekloges-2016/person/create/
Traceback (most recent call last):
  ...
django.core.exceptions.ValidationError: {'start_date': ['date seems not to be correct 1984-06-00']}

...

POST:{'birth_date': '1984-06',
```

`candidates.models.popolo_extra.parse_approximate_date` appends a '-00' to partial dates which [`popolo.behaviors.models.validate_partial_date`](https://github.com/mysociety/django-popolo/blob/fix-migrations-for-python-3/popolo/behaviors/models.py#L30-L45) can't parse.

wfdd:
- Broadly, enable the input and display of day-less dates
- Vary short date format depending on locale (e.g. m/d/y for en-us,
  y/m/d for zh, etc.) and suggest date be input in forms
  in this format if invalid
- Per above, remove support for ``DD_MM_DATE_FORMAT_PREFERRED``
- Remove support for parsing ordinals and prepositions
  (motivation explained in ``DateParser`` docstring)
- Introduce new ``DateCharField`` class to encapsulate
  form-related date operations
- Factor out StrippedCharField into a new module and
  replace EmailField and URLField with their stripepd
  counterparts for completeness

Closes #907.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ python:
   - '3.4'
   - '3.5'
 addons:
+  apt:
+    packages:
+      - libicu-dev
   postgresql: "9.1"
 
 install:
--- a/candidates/date_parsing.py
+++ b/candidates/date_parsing.py
@@ -67,11 +67,10 @@ class DateParser(object):
          ('yMMM', DatePrecision.MONTH)))
     _patterns = OrderedDict(
         (('yyyy-MM-dd', DatePrecision.DAY),   # ISO 8601
-         ('d-M-yyyy', DatePrecision.DAY),     # 10-6-2016
          ('yyyy-MM', DatePrecision.MONTH),
+         ('yyyy', DatePrecision.YEAR),
          ('LLLL y', DatePrecision.MONTH),     # Μάιος, cf. Μαΐου (see http://unicode.org/cldr/trac/ticket/8456)
-         ('LLL y', DatePrecision.MONTH),      # Μάι, cf. Μαΐ
-         ('y', DatePrecision.YEAR)))
+         ('LLL y', DatePrecision.MONTH)))     # Μάι, cf. Μαΐ
     _skeletons_and_patterns = dict(it.chain(_skeletons.items(),
                                             _patterns.items()))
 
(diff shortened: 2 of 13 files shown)


wfdd:
The one remaining failure appears to be unrelated to anything I've done.  I should probably split the changes to compat into a separate commit, if not a separate PR.

struan:
This does add a dependency on the libicu-dev (or equivalent) packaging being installed which I guess should be pointed out somewhere.

struan:
( deleted idiotic comment resulting from me not reading anything :( )

wfdd:
I can't recall if there's anything that remains to be done here - would anybody like to review things?

mhl:
@wfdd I'm sorry, I feel really bad that I didn't give you more feedback on this at the time. I tried to review it properly, but found the review very heavy going, particularly once I started trying to understand ICU and the date stuff in the CLDR.

One thing that in particular wasn't clear to me was how you selected the skeletons and patterns used in `DateParser`, given the huge number of possible formats in the Unicode CLDR. It also wasn't clear to me how we would extend it for new languages.

Overall, my feeling is that the `DateParser` class should be a separate Python package with its own documentation. I think it this could be a very popular package, too. What do you think? Basically, I would be much happier about using it in YNR if the all the complexity was in separately maintained package and codebase - it doesn't seem like something that's core to the project, and it would most likely be useful to many other people.

wfdd:
No worries!  Thanks for taking the time to respond.

It's been some time since I last looked at this, so I'm a little iffy on the details; but I'm sceptical if `DateTimePatternGenerator` is the right tool for the job here.  Trouble is ICU doesn't consistently (?) validate dates and will sometimes happily parse dates of a different spec if they sort of, kinda match a less optimal expanded spec.  Now imagine this happening in an unprioritised loop.  If you look at lines 78 to 85, I did make a start on writing custom validators.  If there's actual potential to this, I agree that it should be developed independently of YNR.

> One thing that in particular wasn't clear to me was how you selected the skeletons and patterns used in DateParser, given the huge number of possible formats in the Unicode CLDR. It also wasn't clear to me how we would extend it for new languages.

Maybe it helps to think of skeletons as the equivalent of `gettext` message IDs and of patterns as their corresponding localised strings.  The idea was that we wouldn't have to cater for languages individually; it'd just work™ because the Unicode folk have already done all the work of pairing skeletons to patterns, by language (and script, and calendar).

Issue #789, opened 2016-04-04
Update party lists to use candidate list form


wfdd:
Party list pages appear to use an older, unsectioned version of the form.

PR #926, opened 2016-05-12
fix party pre-selection in new candidate party list page form

### LINKED ISSUE #855: Party not filled in from 'initial' in ordered party list view new candidate form
The code in `candidates/views/constituencies.py` (line 570) suggests that it should.

struan:
Fixes #855


wfdd:
👍 

Can we combine this with #789 to replace the form with the sectioned one?

wfdd:
Never mind me then.

PR #928, opened 2016-05-13
Replace old candidate form in party list with newer sectioned one

### LINKED ISSUE #789: Update party lists to use candidate list form

wfdd:
Fixes #789.


struan:
Looks good to me.

Issue #929, opened 2016-05-16
Create compact list of winners at top of constituency page


struan:
Brought over from discussions on #869 @wfdd said:

I'm not sure if splitting the lists between elected and unelected candidates is such a good idea. Because, essentially, the post view is a digital representation of the ballot paper, and the ballot paper - probably - should be reproduced as accurately as possible, even after the election has come to a close. I'd opt for repeating the winners compactly under the new 'Elected candidates' heading and restoring the complete list below under 'All candidates'.

PR #930, opened 2016-05-16
856 show elections won on candidate page

struan:
This adds a list of elections won to the top of the person page. There's no limit to how many it shows at the moment. It uses the styles from constituency page which are possibly a bit heavy for this. Possibly once #929 is done it could use that.

currently looks like:

screen shot 2016-05-16 at 16 06 01

### DIFF
--- a/candidates/views/people.py
+++ b/candidates/views/people.py
@@ -121,6 +121,13 @@ class PersonView(TemplateView):
             ).order_by('-election_date')
         else:
             context['elections_to_list'] = elections_by_date
+
+        context['elected_in'] = self.person.memberships.filter(
+            extra__elected=True
+        ).select_related(
+            'post', 'extra__election'
+        ).order_by('-extra__election__election_date')
+
         context['last_candidacy'] = self.person.extra.last_candidacy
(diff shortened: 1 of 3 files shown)


candidates/views/people.py
```diff
@@ -121,6 +121,12 @@ def get_context_data(self, **kwargs):
             ).order_by('-election_date')
         else:
             context['elections_to_list'] = elections_by_date
+
+        context['elected_in'] = self.person.memberships.filter(
+            extra__elected=True
+        ).select_related(
+            'post', 'extra__election'
+        )
```
mhl: It'd probably good to sort these by election date (most recent at the top)

wfdd:
Having something like an 'Elected' badge in the list of candidacies below might also work?  Or do we want this to be prominent?

mhl:
I think it would be good to have "elected" / "non-elected" shown by each candidacy further down as well, for those where we have that information, so the complete history of what they've stood for is a better reference.

struan:
I'm going to leave the candidacies thing for now - I've created #954 for this as I think it needs a bit more thought about the design.

PR #932, opened 2016-05-17
If none of the Twitter details are valid, remove them

mhl:
The candidates_update_twitter_usernames script behaved less than ideally
in the case where none of their Twitter details correspond to a real
Twitter account (typically when they delete their account, or we only
have an old screen name that predates a change of screen name): it would
print out a warning, which would be emailed if the script is run from
a cron job.  This isn't ideal because we never want to be linking to
non-existent Twitter accounts from the site - the action you'd take on
seeing this warning is always going to be to remove or fix the bad data.
(The old values will still be in the version history, after all.)  You
might want to search for a new Twitter account too, but you can still do
that afterwards if the bad data were deleted.

This commit changes the behaviour of the script so this bad Twitter
account data is deleted automatically; it still emits a warning, so those
cases can be investigated (and in some cases new accounts found) by hand
afterwards.

### DIFF
--- a/candidates/management/commands/candidates_update_twitter_usernames.py
+++ b/candidates/management/commands/candidates_update_twitter_usernames.py
@@ -45,11 +45,21 @@ class Command(BaseCommand):
                 .get(contact_type='twitter').value
         except ContactDetail.DoesNotExist:
             screen_name = None
+        except ContactDetail.MultipleObjectsReturned:
+            print(_("WARNING: multiple Twitter screen names found for {name} ({id}), skipping.").format(
+                name=person.name, id=person.id
+            ))
+            return
         try:
             user_id = person.identifiers \
                 .get(scheme='twitter').identifier
         except Identifier.DoesNotExist:
             user_id = None
+        except Identifier.MultipleObjectsReturned:
+            print(_("WARNING: multiple Twitter user IDs found for {name} ({id}), skipping.").format(
+                name=person.name, id=person.id
+            ))
+            return
         # If they have a Twitter user ID, then check to see if we
         # need to update the screen name from that; if so, update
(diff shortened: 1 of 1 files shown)


candidates/management/commands/candidates_update_twitter_usernames.py
```diff
@@ -71,11 +94,12 @@ def handle_person(self, person):
                 person=person, user_id=user_id
```
struan: Might be an idea to update the comment to mention that it deletes if not found.

candidates/management/commands/candidates_update_twitter_usernames.py
```diff
@@ -100,11 +124,12 @@ def handle_person(self, person):
                 person=person, screen_name=screen_name
```
struan: Similarly, updating the comment to mention the delete behaviour probably helpful

candidates/management/commands/candidates_update_twitter_usernames.py
```diff
@@ -71,11 +94,12 @@ def handle_person(self, person):
                 person=person, user_id=user_id
             ))
             if user_id not in self.user_id_to_screen_name:
-                print(_("Warning: user ID {user_id} not found for {person_name}: {person_url}").format(
+                print(_("User ID {user_id} not found for {person_name}; removing it. {person_url}").format(
```
struan: Scanning down a list of these things it seems to me that putting the action that occurred at the start of the line makes it much easier to parse by eye so maybe changing the text here to:

Removing User ID {user_id} for {person_name} as no matching user id found

or something like that.

struan:
Small notes about comments and output format but otherwise looks good to me.

PR #933, opened 2016-05-17
Display candidate elected button on party list page

struan:
This is required in cases where the number of people elected from a
party list exceeds the number that are shown on the constituency page.
Without this the only way to mark people not shown on the constituency
page was to increase the party list length display limit for the
constituency page which is not something most users will be able to do.

### DIFF
--- a/candidates/templates/candidates/ordered-party-list.html
+++ b/candidates/templates/candidates/ordered-party-list.html
@@ -47,45 +47,9 @@
     <h3>{% blocktrans with post_label=post_data.label party_name=party.name %}{{ party_name }} candidates for <a href="{{ post_url }}">{{ post_label }}</a>{% endblocktrans %}
     </h3>
 
-    {% for position_in_list, c, candidate_elected in positions_and_people %}
-      {% if forloop.first %}
-        <ul class="candidate-list">
-      {% endif %}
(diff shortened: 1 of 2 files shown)


candidates/templates/candidates/ordered-party-list.html
```diff
@@ -68,6 +68,17 @@ <h4 class="candidates-list__person__elected-text">{% trans "Winner" %}</h4>
           {% if candidate_list_edits_allowed %}
             {% include 'candidates/_source_confirmation.html' with standing='not-standing' action='candidacy-delete' %}
           {% endif %}
+          {% if DATE_TODAY >= election_data.election_date %}
```
mhl: This template is now substantially similar to `candidates/templates/candidates/ordered-party-list.html` - presumably that partial template could be reused here with a few changes to reduce repetition?

(The ordered-party-list.html file also has some things we'd probably want to leave out, e.g. the old and ugly "Winner" highlighting.)

mhl:
(As we discussed in IRC...) It'd be good to have a test that asserts that on the ordered-party-list page the "See just this party list" text doesn't appear (or something similar), since it might be easy for someone to add `party_lists_in_use` to the view, without realising it'll cause that irrelevant text to appear.

Otherwise, this looks good :+1:

PR #936, opened 2016-05-23
Fix all candidates heading in posts with elected candidates

wfdd:
This reverts to showing the 'Known candidates' heading for the full list
of candidates when hoisting is switched off.


wfdd:
I suppose there was no need to expose the hoist setting globally.

struan:
Looks good to me.

PR #939, opened 2016-05-24
Add confirmation dialog for unselecting winners

### LINKED ISSUE #935: Ask for confirmation when unsetting winners
I've just accidentally reset 13 elected candidates.

struan:
Ask if people want to unset the winners and include count of how many
will be unset to make sure they understand it's all winners

Fixes #935

### DIFF
--- a/candidates/static/js/constituency.js
+++ b/candidates/static/js/constituency.js
@@ -70,9 +70,13 @@ $(function() {
     var enclosingDiv = $(e.target).parent(),
       electedCandidatesDiv=enclosingDiv.find('.candidates__elected'),
       electedCandidates=electedCandidatesDiv.find('li.candidates-list__person'),
+      electedCount=electedCandidates.length,
+      text,
       message;
-    message = interpolate(gettext("Are you sure that you want to unset all %s winners?"),
-        [electedCandidates.length]);
+    text = ngettext("Are you sure that you want to unset one winner?",
+                    "Are you sure that you want to unset all %s winners?",
+                    electedCount);
+    message = interpolate(text, [electedCount]);
     return confirm(message);
   });
 
--- a/candidates/static/js/constituency.js
+++ b/candidates/static/js/constituency.js
@@ -66,4 +66,14 @@ $(function() {
     return confirm(message);
   });
 
+  $('.winner-unset').submit(function(e) {
+    var enclosingDiv = $(e.target).parent(),
+      electedCandidatesDiv=enclosingDiv.find('.candidates__elected'),
+      electedCandidates=electedCandidatesDiv.find('li.candidates-list__person'),
+      message;
+    message = interpolate(gettext("Are you sure that you want to unset all %s winners?"),
+        [electedCandidates.length]);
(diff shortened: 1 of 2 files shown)


candidates/static/js/constituency.js
```diff
@@ -66,4 +66,14 @@ $(function() {
     return confirm(message);
   });
 
+  $('.winner-unset').submit(function(e) {
+    var enclosingDiv = $(e.target).parent(),
+      electedCandidatesDiv=enclosingDiv.find('.candidates__elected'),
+      electedCandidates=electedCandidatesDiv.find('li.candidates-list__person'),
+      message;
+    message = interpolate(gettext("Are you sure that you want to unset all %s winners?"),
```
mhl: I think this [should use `ngettext`](https://docs.djangoproject.com/en/1.9/topics/i18n/translation/#ngettext) instead to get the plural of "winner" right.

mhl: Sorry, I mean "as well", instead of "instead" - it looks like you do `ngettext` then `interpolate`: http://www.djangobook.com/en/2.0/chapter19.html#the-javascript-catalog-view

candidates/static/js/constituency.js
```diff
@@ -70,9 +70,13 @@ $(function() {
     var enclosingDiv = $(e.target).parent(),
       electedCandidatesDiv=enclosingDiv.find('.candidates__elected'),
       electedCandidates=electedCandidatesDiv.find('li.candidates-list__person'),
+      electedCount=electedCandidates.length,
+      text,
       message;
-    message = interpolate(gettext("Are you sure that you want to unset all %s winners?"),
-        [electedCandidates.length]);
+    text = ngettext("Are you sure that you want to unset the winner?",
```
wfdd: this is really minor, but 'the winner' sounds like there can be only one winner, which might not be the case - how about 'Are you sure that you want to unset the one winner?'?

candidates/static/js/constituency.js
```diff
@@ -66,4 +66,18 @@ $(function() {
...
   });
 
+  $('.winner-unset').submit(function(e) {
+    var enclosingDiv = $(e.target).parent(),
+      electedCandidatesDiv=enclosingDiv.find('.candidates__elected'),
+      electedCandidates=electedCandidatesDiv.find('li.candidates-list__person'),
+      electedCount=electedCandidates.length,
+      text,
+      message;
+    text = ngettext("Are you sure that you want to unset the winner?",
```
wfdd: this is really minor, but 'the winner' sounds like there can be only one winner, which might not be the case - how about 'Are you sure that you want to unset the one winner?'?

wfdd: Actually, that's not much better... maybe without the 'the'.

mhl:
Looks good :+1: Just as a note, after the merge we'll need to do the steps in https://github.com/mysociety/yournextrepresentative/blob/master/docs/transifex.md - in particular not forgetting the `makemessages -d djangojs` step.

Issue #944, opened 2016-05-25
design a basic (non-Django admin) admin interface


mhl:
(We probably need a better name for that...)

I'm slightly in two minds about this. The user need is for a more user friendly interface for configuring and maintaining a YNR site without needing to edit files on the server or be a Django developer. We could attempt to customize the Django admin interface to do this, but we suspect that this will turn out to be increasingly difficult as we try to improve the user experience. (Anecdotally, at Djangocon EU I heard a few people say that their projects had run into problems by trying to do too much in the Django admin interface, for what that's worth.)

So, the idea of this issue is that rather than putting a lot of work into customizing the existing Django admin interface, we should start building a new interface (like that of FMS or Alaveteli) for common settings and administration tasks. This might ultimately include, for example:

* A "settings" tab for customizing basic site settings #943 
* A "parties" tab for adding, deleting and updating parties (including by update from EveryPolitician / Wikdata / CSV upload / Popolo JSON upload)
* An "elections" tab for adding, deleting and updating parties (including by CSV / Popolo JSON upload)
* An "elected bodies" tab for 
* User management

jf1:
The people doing the work should make such decisions but fwiw, having only
seen a little of the Django admin interface, I think this is a sensible and
worthwhile approach.

Once the additional code and data fields are in place, sh/could a "parties"
tab allow/enable selection of a "default party description" to deal with
https://github.com/mysociety/yournextrepresentative/issues/823

PR #951, opened 2016-05-31
Move site settings into database and provide UI

### LINKED ISSUE #943: create a settings table for site-specific configuration in the database
In order to make it easier to configure a new YourNextRepresentative instance without having to edit files, we'd like to move some of the settings in `conf/general.yml` into a new settings table.

For example, we would want this to include these settings:

* The country that this site is for
* The default language and timezone
* `DATE_FORMAT`
* `SUPPORT_EMAIL`
* `GOOGLE_ANALYTICS_ACCOUNT`
* `TWITTER_USERNAME`
* `RESTRICT_RENAMES`
* `NEW_ACCOUNTS_ALLOWED`
* `EDITS_ALLOWED`
* `HOIST_ELECTED_CANDIDATES`
* `DD_MM_DATE_FORMAT_PREFERRED` *

... and possibly others.

* https://github.com/mysociety/yournextrepresentative/pull/918 would remove this setting anyway

Some questions to address here:

* Should these settings be related to a Site from the sites framework via a foreign key? (Probably.)
* Is there any point in using an existing app to manage this? (Probably not.)

struan:
Move most of the settings for setting up a site into the database and provide an interface to edit them as part of the plan to make site setup easier.

Does not handle country, language or timezone because of loading complexities.

Also does not handle AREAS_TO_ALWAYS_RETURN as this is a bit more complicated than a simple setting.

Uses django-usersettings2 as a basis.

Fixes #943 
Also partially handles #944 in that there is an interface for the basic settings.

### DIFF
--- a/candidates/admin.py
+++ b/candidates/admin.py
@@ -5,9 +5,12 @@ from django.contrib import admin
 from django.core.urlresolvers import reverse
 from django.forms import ModelForm
 
+from usersettings.admin import SettingsAdmin
+
 from .models import (
     LoggedAction, PartySet, ExtraField, PersonExtraFieldValue,
-    SimplePopoloField, ComplexPopoloField, PostExtraElection
+    SimplePopoloField, ComplexPopoloField, PostExtraElection,
+    SiteSettings
 )
 
 
@@ -81,3 +84,35 @@ class PostExtraElectionAdmin(admin.ModelAdmin):
 class ComplexPopoloFieldAdmin(admin.ModelAdmin):
     list_display = ['name', 'label', 'order']
     ordering = ('order',)
+
+@admin.register(SiteSettings)
+class SiteSettingsAdmin(SettingsAdmin):
+    fieldsets = (
+        ("Site owner's details", {
+            'fields': ('SITE_OWNER', 'SITE_OWNER_URL', 'COPYRIGHT_HOLDER',
+                'TWITTER_USERNAME')
+        }),
+        ('Email addresses', {
+            'fields': ('SUPPORT_EMAIL', 'DEFAULT_FROM_EMAIL', 'SERVER_EMAIL')
+        }),
+        ('Localization', {
+            'fields': ('DATE_FORMAT', 'DD_MM_DATE_FORMAT_PREFERRED')
+        }),
+        ('Areas', {
+            'fields': ('MAPIT_BASE_URL',)
+        }),
+        ('Google Analytics', {
+            'fields': ('GOOGLE_ANALYTICS_ACCOUNT', 'USE_UNIVERSAL_ANALYTICS')
+        }),
+        ('Display options', {
+            'fields': ('HOIST_ELECTED_CANDIDATES',
+                       'CANDIDATES_REQUIRED_FOR_WEIGHTED_PARTY_LIST')
+        }),
+        ('Editing and account options', {
+            'fields': ('NEW_ACCOUNTS_ALLOWED', 'RESTRICT_RENAMES',
+                       'EDITS_ALLOWED')
+        }),
+        ('Other', {
+            'fields': ('TWITTER_APP_ONLY_BEARER_TOKEN', 'IMAGE_PROXY_URL')
+        })
+    )
--- a/candidates/forms.py
+++ b/candidates/forms.py
@@ -15,7 +15,8 @@ from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 
 from candidates.models import (
-    PartySet, parse_approximate_date, ExtraField, SimplePopoloField, ComplexPopoloField
+    PartySet, parse_approximate_date, ExtraField, SimplePopoloField,
+    ComplexPopoloField, SiteSettings
 )
 from popolo.models import Organization, OtherName, Post
 from .twitter_api import get_twitter_user_id, TwitterAPITokenMissing
@@ -526,3 +527,28 @@ class OtherNameForm(forms.ModelForm):
         ),
         max_length=512,
     )
+
+
+class SettingsForm(forms.ModelForm):
+    class Meta:
+        model = SiteSettings
+        fields = (
+            'SITE_OWNER',
+            'SITE_OWNER_URL',
+            'COPYRIGHT_HOLDER',
+            'TWITTER_USERNAME',
+            'SUPPORT_EMAIL',
+            'DEFAULT_FROM_EMAIL',
+            'SERVER_EMAIL',
+            'DATE_FORMAT',
+            'DD_MM_DATE_FORMAT_PREFERRED',
+            'MAPIT_BASE_URL',
+            'GOOGLE_ANALYTICS_ACCOUNT',
+            'USE_UNIVERSAL_ANALYTICS',
+            'NEW_ACCOUNTS_ALLOWED',
+            'HOIST_ELECTED_CANDIDATES',
+            'RESTRICT_RENAMES',
+            'EDITS_ALLOWED',
+            'CANDIDATES_REQUIRED_FOR_WEIGHTED_PARTY_LIST',
+            'TWITTER_APP_ONLY_BEARER_TOKEN',
+            'IMAGE_PROXY_URL')
--- a/candidates/migrations/0033_migrate_settings_to_db.py
+++ b/candidates/migrations/0033_migrate_settings_to_db.py
@@ -1,0 +1,85 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import importlib
+
+from django.conf import settings
+from django.db import migrations
+
+from mysite.settings.conf import get_conf
+
+def get_election_app_setting(setting_name):
+    election_app = settings.ELECTION_APP
+    if not election_app:
+        return None
+
+    election_app_fully_qualified = 'elections.' + election_app
+    election_settings_module = election_app_fully_qualified + '.settings'
+    elections_module = importlib.import_module(election_settings_module)
+    try:
+        setting = getattr(elections_module, setting_name)
+    except AttributeError:
+        setting = ''
+
+    return setting
+
+def migrate_settings(apps, schema_editor):
+    """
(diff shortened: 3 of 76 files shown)


mhl:
There's a typo in the commit message of 46dc863dc9e4669fa19 - "Users" should be "Uses"

candidates/migrations/0033_migrate_settings_to_db.py
```diff
@@ -0,0 +1,85 @@
...
+            SERVER_EMAIL=old_settings['SERVER_EMAIL'],
+            DEFAULT_FROM_EMAIL=old_settings['DEFAULT_FROM_EMAIL'],
+            SUPPORT_EMAIL=old_settings['SUPPORT_EMAIL'],
+            SITE_OWNER=get_election_app_setting('SITE_OWNER'),
+            SITE_OWNER_URL=get_election_app_setting('SITE_OWNER_URL'),
+            COPYRIGHT_HOLDER=get_election_app_setting('COPYRIGHT_HOLDER'),
+            MAPIT_BASE_URL=get_election_app_setting('MAPIT_BASE_URL'),
+            IMAGE_PROXY_URL=get_election_app_setting('IMAGE_PROXY_URL'),
+            GOOGLE_ANALYTICS_ACCOUNT=old_settings['GOOGLE_ANALYTICS_ACCOUNT'],
+            USE_UNIVERSAL_ANALYTICS=old_settings['USE_UNIVERSAL_ANALYTICS'],
```
mhl: `USE_UNIVERSAL_ANALYTICS` is optional (defaulting to `True`)

candidates/migrations/0033_migrate_settings_to_db.py
```diff
@@ -0,0 +1,85 @@
...
+    user = superusers.first()
+
+    old_settings = get_conf('general.yml')
+
+    if user and get_election_app_setting('SITE_OWNER'):
+        SiteSettings.objects.using(db_alias).create(
+            site_id=1,
+            user_id=user.id,
+            # TODO: check default
+            DATE_FORMAT=old_settings.get('DATE_FORMAT', ''),
```
mhl: I can't see `DATE_FORMAT` in the history in git, so I'm not sure where this is from?

candidates/migrations/0033_migrate_settings_to_db.py
```diff
@@ -0,0 +1,85 @@
...
+            SITE_OWNER=get_election_app_setting('SITE_OWNER'),
+            SITE_OWNER_URL=get_election_app_setting('SITE_OWNER_URL'),
+            COPYRIGHT_HOLDER=get_election_app_setting('COPYRIGHT_HOLDER'),
+            MAPIT_BASE_URL=get_election_app_setting('MAPIT_BASE_URL'),
+            IMAGE_PROXY_URL=get_election_app_setting('IMAGE_PROXY_URL'),
+            GOOGLE_ANALYTICS_ACCOUNT=old_settings['GOOGLE_ANALYTICS_ACCOUNT'],
+            USE_UNIVERSAL_ANALYTICS=old_settings['USE_UNIVERSAL_ANALYTICS'],
+            TWITTER_USERNAME=old_settings.get('TWITTER_USERNAME', ''),
+            TWITTER_APP_ONLY_BEARER_TOKEN=old_settings.get('TWITTER_APP_ONLY_BEARER_TOKEN', ''),
+            RESTRICT_RENAMES=old_settings['RESTRICT_RENAMES'],
```
mhl: `RESTRICT_RENAMES` is optional too (defaulting to `None`, currently). In fact, I'd be inclined to replace `[]` with `get` in all these cases just so it's more tolerant if someone's configuration is bad at the point they run the migration.

candidates/admin.py
```diff
@@ -81,3 +84,34 @@ class PostExtraElectionAdmin(admin.ModelAdmin):
...
+            'fields': ('DATE_FORMAT', 'DD_MM_DATE_FORMAT_PREFERRED')
+        }),
+        ('Areas', {
+            'fields': ('MAPIT_BASE_URL',)
+        }),
+        ('Google Analytics', {
+            'fields': ('GOOGLE_ANALYTICS_ACCOUNT', 'USE_UNIVERSAL_ANALYTICS')
+        }),
+        ('Display and editing options', {
+            'description': 'How candidates are displayed and if eding is allowed',
```
mhl: There's a typo here: "eding" -  I'd probably split this into two groups, though - one for the two display options, and the other for those that restrict what users can do.

candidates/admin.py
```diff
@@ -81,3 +84,34 @@ class PostExtraElectionAdmin(admin.ModelAdmin):
...
+            'fields': ('NEW_ACCOUNTS_ALLOWED', 'RESTRICT_RENAMES', 'EDITS_ALLOWED',
+                'HOIST_ELECTED_CANDIDATES', 'CANDIDATES_REQUIRED_FOR_WEIGHTED_PARTY_LIST')
+
+        }),
+        ('Other', {
+            'description': 'How candidates are displayed and if eding is allowed',
```
mhl: This needs a new description.

candidates/admin.py
```diff
@@ -81,3 +84,34 @@ class PostExtraElectionAdmin(admin.ModelAdmin):
...
     list_display = ['name', 'label', 'order']
     ordering = ('order',)
+
+@admin.register(SiteSettings)
+class SiteSettingsAdmin(SettingsAdmin):
+    fieldsets = (
+        ("Site owner's details", {
+            'fields': ('SITE_OWNER', 'SITE_OWNER_URL', 'COPYRIGHT_HOLDER',
+                'TWITTER_USERNAME')
+        }),
```
mhl: Grouping these is nice :+1:

candidates/forms.py
```diff
@@ -526,3 +527,19 @@ class Meta:
...
+            'SITE_OWNER', 'SITE_OWNER_URL', 'COPYRIGHT_HOLDER',
+            'TWITTER_USERNAME',
+            'SUPPORT_EMAIL', 'DEFAULT_FROM_EMAIL', 'SERVER_EMAIL',
+            'DATE_FORMAT', 'DD_MM_DATE_FORMAT_PREFERRED',
+            'MAPIT_BASE_URL',
+            'GOOGLE_ANALYTICS_ACCOUNT', 'USE_UNIVERSAL_ANALYTICS',
+            'NEW_ACCOUNTS_ALLOWED', 'HOIST_ELECTED_CANDIDATES',
+            'RESTRICT_RENAMES', 'EDITS_ALLOWED',
+            'CANDIDATES_REQUIRED_FOR_WEIGHTED_PARTY_LIST',
+            'TWITTER_APP_ONLY_BEARER_TOKEN', 'IMAGE_PROXY_URL')
```
mhl: It's a tiny thing, but I think it's a bit clearer just to have one per line here.

candidates/templates/candidates/settings.html
```diff
@@ -0,0 +1,205 @@
...
+
+  <div class="form-item {% if form.USE_UNIVERSAL_ANALYTICS.errors %}form-item--errors{% endif %}">
+    <p>
+      {{ form.USE_UNIVERSAL_ANALYTICS }}
+      - <label for="id_USE_UNIVERSAL_ANALYTICS">{% trans "Site uses Universal Google analytics" %}</label>
+    </p>
+    {{ form.USE_UNIVERSAL_ANALYTICS.errors }}
+  </div>
+
+  <h2>{% trans "Display and editing options" %}</h2>
```
mhl: As with the admin, I'd probably divide this into two groups

candidates/views/settings.py
```diff
@@ -24,9 +27,30 @@ def get_context_data(self, **kwargs):
...
+                        old_value=getattr(settings, field),
+                        new_value=form[field].value()
+                ) + "\n"
             setattr(settings, field, form[field].value())
         settings.user = self.request.user
         settings.save()
 
+        request = self.request
+
+        if note != '':
```
mhl: I think just `if note` is marginally more idiomatic.

candidates/tests/test_settings.py
```diff
@@ -0,0 +1,100 @@
...
+            'jane@example.com',
+            'alsonotagoodpassword',
+        )
+        self.test_setter.terms_agreement.assigned_to_dc = True
+        self.test_setter.terms_agreement.save()
+        self.test_setter.groups.add(
+            Group.objects.get(name=EDIT_SETTINGS_GROUP_NAME)
+        )
+
+    def test_photo_review_view_unprivileged(self):
```
mhl: I think this and the next method have names copy-and-pasted from elsewhere...

candidates/tests/settings.py
```diff
@@ -17,10 +17,6 @@ def setUp(self):
         self.sitesettings = SiteSettings.objects.create(
             site_id=1,
             user_id=self.settings_user.id,
-            # COUNTRY=settings.COUNTRY,
-            LANGUAGE_CODE='en-gb',
-            TIME_ZONE='Europe/London',
-            # DATE_FORMAT=settings.DATE_FORMAT,
```
mhl: I think this change should probably be split and squashed into 1aa02cfae2d13cbe

mhl:
Nice work! I've left a few very minor comments, but :+1: generally.

The concern I mentioned in IRC probably doesn't matter much, but to explain further: django-usersettings2 ties into the sites framework, so you can have different settings for each site. (I think we discussed this and thought it would be a Good Thing.) In the future, however, if we implement hosting multiple YNR instances as different subdomains, with a single database, we'd probably do that with [django-subdomain-instances](https://github.com/mysociety/django-subdomain-instances), like SayIt - in that case you would want the settings to be associated with an instance rather than a site. I think we can deal with that issue when we come to it, though; we could fork django-usersettings2 and associate its settings with an instance, or add optional creation of a site per instance in django-subdomain-instances. (Or we might use a completely different strategy for allowing easier instance creation - e.g. using a PaaS like Heroku.)

mhl:
Oh, and there's a test failure still.

PR #952, opened 2016-06-01
Allow election AreaType to be blank in forms/admin

### LINKED ISSUE #942: selecting an area type shouldn't be mandatory when editing election in the admin interface
And then this bit of documentation can be removed: https://github.com/mysociety/yournextrepresentative/blame/72a3eddfbce718265054415c8841502dead98831/docs/new-instance.md#L122-L133

struan:
Even though AreaType is a ManyToMany relationship and therefore option
unless blank=True is set on the field then the admin interface will
insist that you associate an election with an AreaType. We might not
want to do this if we are not using AreaType lookups, e.g. a simple YNR
instance which has no geolocation area lookup.

Fixes #942

### DIFF
--- a/elections/migrations/0013_optional_election_area_type.py
+++ b/elections/migrations/0013_optional_election_area_type.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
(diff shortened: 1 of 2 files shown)


mhl:
:+1: Looks good - are you planning to do a single update of the documentation after all these, or should the update be in this commit too?

struan:
I think a single update to documentation would be preferable otherwise I suspect lots of time on rewriting the same sections for each change.

PR #966, opened 2016-08-30
Fix lack of validation errors from /settings and a caching bug

mhl:
No validation errors were being shown when trying to update settings at
/settings, because the form had the settings object instance set on it
in the wrong way: in get_context_data rather than via get_kwargs.

This commit fixes that, but this also exposed a bug where the cached
setting object was being updated when checking the validity of the
fields. This commit fixes that as well, otherwise the switch to
using get_form_kwargs causes test failures because of this bad
caching.


chrismytton:
Looks good to me 👍

PR #968, opened 2016-11-30
Add a wizard to add a new candidacy for a person

mhl:
The motivation of this pull request is to deal with the problem that the person edit page included post and party selects for every current election in the database - this might be repeated for hundreds of elections, which makes the page very slow to load and crashes some browsers. (See https://github.com/mysociety/yournextrepresentative/pull/740 for more details.)

On the DemocracyClub fork of YNR, this problem has been dealt with by using Javascript to dynamically add those form elements. However, in this upstream versions, so far, all of the site works without Javascript, on the principle that progressive enhancement is a Good Thing, so I was reluctant to merge those changes as-is. This pull request adds a method which is not Javascript-dependent to do the same thing, and we can use this as a base to apply the DC Javascript enhancement in the future. (Which I'm intending to do.)

The person edit page still uses lots of queries in its generation, which is bad, especially since there are fixes for those issues that I made in the DC fork; however, those fixes will apply most easily after the Javascript enhancement has been merged, so my plan is to get this merged, apply the Javascript enhancement of adding new candidacies, and then apply the fixes for excessive queries.

I guess @symroe or @struan might be best placed to review this, if either of you have time? n.b. @symroe, we discussed recently that the formtools wizard examples suggest you need to specify `condition_dict` as a keyword parameter to `as_view`, but it works fine when specified as a class attribute instead, as in this example.

add-candidacy-link
wizard-election-step

### DIFF
--- a/candidates/forms.py
+++ b/candidates/forms.py
@@ -12,11 +12,12 @@ from django import forms, VERSION as django_version
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.core.exceptions import ValidationError
+from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
 
 from candidates.models import (
     PartySet, parse_approximate_date, ExtraField, SimplePopoloField,
-    ComplexPopoloField, SiteSettings
+    ComplexPopoloField, SiteSettings, PostExtra
 )
 from popolo.models import Organization, OtherName, Post
 from .twitter_api import get_twitter_user_id, TwitterAPITokenMissing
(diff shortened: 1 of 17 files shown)


candidates/views/candidacies.py
```diff
@@ -128,3 +132,119 @@ def get_context_data(self, **kwargs):
...
+        party_id = form_party.cleaned_data['party']
+        party = get_object_or_404(Organization, pk=party_id)
+        party_list_position = form_party.cleaned_data.get('party_list_position')
+        raise_if_locked(self.request, post)
+        change_metadata = get_change_metadata(
+            self.request, form_source.cleaned_data['source']
+        )
+        with transaction.atomic():
+            person = get_object_or_404(Person, id=kwargs['person_id'])
+            LoggedAction.objects.create(
```
mhl: The context is only prepared for GET requests, and `done` is called when handling a POST request, but you're quite right that it's strange to have two bits of code that looks up the person...  I've pushed a fixup that looks it up once by when overriding `dispatch`.

candidates/forms.py
```diff
@@ -552,3 +553,103 @@ class Meta:
             'CANDIDATES_REQUIRED_FOR_WEIGHTED_PARTY_LIST',
             'TWITTER_APP_ONLY_BEARER_TOKEN',
             'IMAGE_PROXY_URL')
+
+
+class AddCandidacyPickElectionForm(forms.Form):
+
+    election = forms.ModelChoiceField(
+        queryset=Election.objects.order_by('-election_date', 'name'))
```
symroe: I think this should filter on `current=True` too.

symroe: Or is the idea that people should be able to add non-current elections? I think this could be a bit confusing when there are 1000s of election objects in the database.

mhl: @symroe Yeah, my idea was that you need to have some way to add candidacies for historic elections in order to correct mistakes, or complete old data. But you're quite right that it'll be confusing in the more typical case. I've make a pull request that makes that restricts to `current=True` by default, but links to a version of the picker with all elections: https://github.com/mysociety/yournextrepresentative/pull/969

PR #986, opened 2017-07-27
Kenya styling

zarino:
Applies the logo and header treatment from Mzalendo, and uses Mzalendo colours elsewhere.

Outstanding questions:

* Does the site have a name? (Right now, it looks like the site’s name is just "Mzalendo" which could confuse people)
* How do we want to handle a "Back to Mzalendo" link?
* Do we want to rewrite the homepage text content?

screen shot 2017-07-27 at 11 49 52

screen shot 2017-07-27 at 11 50 04


mhl:
Thanks, @zarino - this all looks good. I think it's probably best if we merge and deploy this updated styling first, and then address those questions with Jessica.

(So that you know: we're also experimenting with changing the free text address entry with a drop-down to select a ward directly, and Jessica has said we should remove the geolocation link (which I think is a shame), so we'll have to make some additional changes, but I think those will be very minimal and probably shouldn't need to take up any more of your time. Changing the title text image will, of course.)

mhl:
n.b. The test failures are unrelated and should be fixed separately.

PR #990, opened 2017-08-01
KE: Replace address search box with county-level dropdown

jacksonj04:
Replaces the search box on the home page with a dropdown of counties.


mhl:
A couple of quick notes on this since I'm away for a few days:

* We still have to decide if this is the right thing to do - I would still prefer to do a quick series of prototypes, as we discussed at the end of last week, even on a static page or on jsfiddle with different options for selecting areas from a drop-down, including variants which included constituencies or even wards, which we can then get people from Mzalendo to try.

* This should really have a test or two in `elections/kenya/tests.py` (the test failures at the moment should be fixed by rebasing over master now)

PR #993, opened 2017-08-04
KE: Change label for location search box on front page

jacksonj04:
Adds a custom form (and necessary URL and view) to modify the label on the front page of YNR Kenya to "Enter your county or constituency"

screen shot 2017-08-04 at 14 14 29

### DIFF
--- a/elections/kenya/forms.py
+++ b/elections/kenya/forms.py
@@ -2,3 +2,24 @@
 # those definitions here.
 
 from __future__ import unicode_literals
+from django import forms
+
+from candidates.models.address import check_address
+
+from candidates.forms import StrippedCharField
+
+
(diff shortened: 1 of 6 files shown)


elections/kenya/forms.py
```diff
@@ -2,3 +2,44 @@
...
+            super(StrippedCharField, self).__init__(max_length, min_length,
+                                                    *args, **kwargs)
+
+        def to_python(self, value):
+            value = super(StrippedCharField, self).to_python(value)
+            if self.strip:
+                value = value.strip()
+            return value
+else:
+    StrippedCharField = forms.CharField
```
mhl: There's no need to duplicate this, I think - you could import it from `candidates.forms` instead.

